Issue #27, opened 2022-03-29
Missing standard/basic training features


riley-mld:
At the moment we're missing some basic training stuff like learning rate schedule. We are also missing some standard RL practices like observation clipping and reward clipping an option for terminal masking. They all should be fairly simple to implement and hopefully should give us small performance gains.

PR #29, opened 2022-03-30
Added more logging

singhblom:
I haven't made sure the tests work yet but thought I'd drop this here so people can have a look. This addresses #27


tgolsson:
Poke @singhblom - there's some interesting things in here I'd like to get in w/ typing, interfaces etc. Better now than in another half a year ;-)

singhblom:
We decided to go in a different direction with the logging.

PR #37, opened 2022-04-11
try fix installs on MacOS

tgolsson:
To ensure functionality on MacOS this PR does the following:

* replaces torch with a CPU-variant
* replaces box2d with box2d-py, which is OpenAI's fork


tgolsson:
Will take another stab at this but since this predates PDM I'll redo it.

PR #69, opened 2022-11-17
Optimize CI performance with preinstalls and dedicated CPU

tgolsson:
Noticied when working on $OTHER_REPO that there's some wonkiness with how we install gcloud and setup our CI tasks that leads to subpar performance. This change should have very little impact on regular running but when the builds cluster is under heavy load we'll not degrade as heavily as we have in the past.

In practice we have ended up running on "any" resource available. I noticed this because we had crammed a bunch of pytest runners into a single available CPU slot, causing some 10-second checks to run for 10+ minutes.

### DIFF
--- a/.buildkite/Dockerfile
+++ b/.buildkite/Dockerfile
@@ -10,6 +10,12 @@ RUN apt-get -y update && \
     apt-get -y install curl build-essential gcc git && \
     apt-get -y clean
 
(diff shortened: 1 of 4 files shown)


singhblom (approved):
I have no strong opinions here. Looks good to me, faster is better, preinstall as much as you can!

PR #78, opened 2023-02-17
add memory warmup callback waiter

tgolsson:
Not sure how to test this easily; but avoids putting this into the trainer. Required since we can't easily do warm-up steps when communicating over network.

### DIFF
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -175,16 +175,17 @@ class MemoryLoader:
 
 class MemoryWarmup(Callback):
     """A blocker to ensure memory has data.
-    
-    This ensures the memory has enough data when training starts, as the memory 
-    will panic otherwise. This is useful if you use an async data generator. 
-    
-    If you do not use an async data generator this will deadlock your training 
-    loop and prevent progress. 
+
+    This ensures the memory has enough data when training starts, as the memory
+    will panic otherwise. This is useful if you use an async data generator.
+
+    If you do not use an async data generator this can deadlock your training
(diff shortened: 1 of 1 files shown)


emote/memory/memory.py
```diff
@@ -169,3 +171,15 @@ def __iter__(self):
 
             data[self.size_key] = self.rollout_count * self.rollout_length
             yield {self.data_group: data, self.size_key: data[self.size_key]}
+
+
+class MemoryWaiter(Callback):
```
singhblom, line 176: LGTM but could we come up with a different name? This didn't do what I expected it to do. `MemoryReady` maybe? I'm not sure.

tgolsson: `MemoryWarmupWaiter`?

singhblom: Waiter sounds like it will bring me water and get me the check. But sure. I'd be fine with just `MemoryWarmup`.

emote/memory/memory.py
```diff
@@ -169,3 +171,24 @@ def __iter__(self):
...
+    
+    This ensures the memory has enough data when training starts, as the memory 
+    will panic otherwise. This is useful if you use an async data generator. 
+    
+    If you do not use an async data generator this will deadlock your training 
+    loop and prevent progress. 
+    """
+    
+    def __init__(self, loader: MemoryLoader):
+        super().__init__()
```
singhblom, line 187: Maybe set self._order to 100 or something. Then this will run after the GymCollector and we won't deadlock in practice.

PR #80, opened 2023-02-17
enable exporting ONNX policies for Gaussian MLPs

tgolsson:
This is approximately about what we talked about over Slack; but I don't think it's super-elegant. I'm wondering whether leaning on `torch.jit.trace` so we can put it into the main class would make more sense. That removes the requirement of the behavior being in `forward`.


tgolsson:
I believe this will have to partially go on the agent proxy as that is where we deal with the dictionary keys. And that leads to using explicit tracing.

emote/sac.py
```diff
@@ -295,6 +300,71 @@ def load_state_dict(self, state_dict: Dict[str, Any]):
...
+        self._counter += 1
+
+        if (self._counter % self._cycle) == 0:
+            self._end_cycle()
+            self._counter = 0
+
+        return self._inner(observations)
+
+
+class AgentProxyOnnxExporter(OnnxGeneratorMixin, OnnxStorage, LoggingProxyWrapper):
```
singhblom, line 323: I don't understand what we gain from having the Onnx-things being mixins. They are not used anywhere else. Why not just let them be private methods?

emote/nn/gaussian_policy.py
```diff
@@ -43,13 +44,13 @@ def __init__(
...
     ):
-        super(GaussianPolicyHead, self).__init__()
+        super().__init__()
         self.action_dim = action_dim
         self.hidden_dim = hidden_dim
         self.mean = nn.Linear(hidden_dim, action_dim)
         self.log_std = nn.Linear(hidden_dim, action_dim)
 
-    def forward(self, x: Tensor) -> Tuple[Tensor, float]:
+    def forward(self, x: Tensor, epsilon: Tensor | None = None) -> Tuple[Tensor]:
```
singhblom, line 53: This return type is maybe false.

emote/extra/onnx_storage.py
```diff
@@ -0,0 +1,164 @@
...
+        # It takes about 20ms to calculate
+        self.version_tag = _get_version()
+        self.scopes = BlockTimers()
+
+    @abstractmethod
+    def generate_onnx(self) -> ModelProto:
+        """Generate a ModelProto for the policy."""
+        ...
+
+    def process_pending_exports(self):
```
singhblom, line 89: This is only used in _export. Maybe inline it?

emote/extra/onnx_storage.py
```diff
@@ -0,0 +1,164 @@
...
+        item = QueuedExport(metadata)
+        self.queued_exports.put(item)
+        if sync:
+            # This will cause block_until_complete to never block
+            # because the work will have been completed already.
+            self.process_pending_exports()
+
+        return item.block_until_complete()
+
+    def export_threadsafe(self, metadata=None) -> StorageItem:
```
singhblom, line 137: This is never used. Is this whole API necessary when we in the code only follow the non-threadsafe path? We could cut a whole lot of code if this is not needed.

emote/extra/onnx_storage.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Same as `export`, but it can be called in threads other than the main thread.
+        This method relies on the main thread calling `process_pending_exports` from time to time.
+        You cannot call this method from the main thread. It will block indefinitely.
+        """
+        return self._export(metadata, False)
+
+    def export(self, metadata=None) -> StorageItem:
```
singhblom, line 145: If this whole mixin is a part of the `AgentProxyOnnxExporterWrapper` ™️  then we could just have export inlined in `_end_cycle`.

emote/extra/onnx_storage.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Serializes a model to onnx and saves it to disk.
+        This must only be called from the main thread.
+        That is, the thread which has ownership over the model and that modifies it.
+        This is usually the thread that has the training loop.
+        """
+        return self._export(metadata, True)
+
+    def delete(self, handle: StorageItemHandle) -> bool:
```
singhblom, line 154: These CRUD operations are never used. We can add them back if they are ever needed.

emote/extra/onnx_storage.py
```diff
@@ -0,0 +1,164 @@
...
+
+from google.protobuf import text_format
+from onnx import ModelProto, helper
+
+from emote.utils.timed_call import BlockTimers
+
+from .crud_storage import CRUDStorage, StorageItem, StorageItemHandle
+
+
+class QueuedExport:
```
singhblom, line 20: This whole class can be cut if we don't care about threadsafe.

emote/extra/threading.py
```diff
@@ -0,0 +1,87 @@
+import threading
```
singhblom, line 1: There's already a LockedResource in `emote/utils` that looks very similar.

emote/extra/onnx_storage.py
tgolsson: Ah good catch. No. It has to be called from a callback somewhere :-)

emote/extra/threading.py
tgolsson: Ah, it's likely the same.

singhblom: This still shows up when I review it. Did you remove it? I feel like a rebase has gone wrong or something here.

emote/onnx_export.py
```diff
@@ -0,0 +1,66 @@
...
+        input_shapes.append(
+            (
+                "epsilon",
+                (*spaces.actions.shape,),
+            )
+        )
+
+    output_shapes = [("actions", (*spaces.actions.shape,))]
+
+    storage = OnnxExporter(
```
singhblom, line 56: If onnx exporter is a storage, name it `OnnxStorage`. If it's an exporter, say `exporter = OnnxExporter(...)`

emote/extra/onnx_storage.py
```diff
@@ -0,0 +1,208 @@
...
+        with open(path, "wb") as f:
+            f.write(message.SerializeToString())
+
+
+class OnnxExporter:
+    """Handles onnx exports of a ML policy.
+
+    Call `export` whenever you want to save an onnx version of the current model.
+
+    This is a mixin class that can be used with diamond inheritance to
```
singhblom, line 64: No longer true right?

emote/onnx_export.py
```diff
@@ -0,0 +1,66 @@
...
+        agent_proxy.policy,
+        directory,
+        input_shapes,
+        output_shapes,
+    )
+
+    return (
+        storage,
+        OnnxExporterCallback(storage, export_interval),
```
singhblom, line 65: Why return a bundle and not just return the OnnxExporter and then create the callback outside?

singhblom (commented):
I think this is getting really nice, so please forgive my requesting one more round of thinking on this one. In addition to the comments below I also have a feeling we can simplify the difference between the onnx policy and the standard policy to avoid giving two different return types for the policy. However, that might have to come at a later point. Have a look at the feedback and let me know what you think. Excellent work.

emote/callbacks.py
```diff
@@ -12,6 +12,7 @@
 from torch.utils.tensorboard import SummaryWriter
 
 from .callback import Callback
+from .extra.threading import LockedResource
```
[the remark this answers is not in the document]
tgolsson: I thought I removed that 😕 Will do!

emote/extra/onnx_storage.py
tgolsson: Good point.

emote/onnx_export.py
tgolsson: Because the ONNX callback is required for proper functioning. :-)

emote/extra/crud_storage.py
```diff
@@ -0,0 +1,213 @@
...
+        handle = StorageItemHandle(self._filename_counter.increment())
+        item = StorageItem(timestamp=utcdate, filepath=filepath, handle=handle)
+
+        with self._items as items:
+            items.append(item)
+        return item
+
+    def create_with_saver(self, saver: Callable[[str], None]) -> StorageItem[T]:
+        """
+        Creates a new file by saving it via the provided function.
```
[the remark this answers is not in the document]
tgolsson: It does in the class docstring. :-)

tgolsson:
> simplify the difference between the onnx policy and the standard policy to avoid giving two different return types for the policy

Yeah; I did one version (like we talked about) with nn.Sequential but it IMO didn't feel as smooth. I'll take another round tomorrow at that.

emote/nn/gaussian_policy.py
tgolsson: Oh hmm, yeah.

tests/test_htm.py
```diff
@@ -57,10 +80,18 @@ def test_htm():
 
     q1 = QNet(2, 1)
     q2 = QNet(2, 1)
-    policy = Policy(2, 1)
+    policy = Policy(2, 1, [N_HIDDEN, N_HIDDEN])
     ln_alpha = torch.tensor(1.0, requires_grad=True)
     agent_proxy = FeatureAgentProxy(policy, device)
 
+    storage, callback = make_onnx_exporter_bundle(
```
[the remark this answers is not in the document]
tgolsson: Yeah; hm.

emote/onnx_export.py
tgolsson: Merged these two into one.

emote/extra/onnx_exporter.py
```diff
@@ -0,0 +1,246 @@
...
+        os.makedirs(dir_name, exist_ok=True)
+    if as_text:
+        with open(path, "w") as f:
+            f.write(text_format.MessageToString(message))
+    else:
+        with open(path, "wb") as f:
+            f.write(message.SerializeToString())
+
+
+class OnnxExporter(Callback):
```
singhblom, line 62: Nice!

emote/extra/onnx_exporter.py
```diff
@@ -0,0 +1,246 @@
...
+                model_proto.model_version = model_version
+                model_proto.doc_string = "exported via Emote checkpointer"
+
+                if metadata is not None:
+                    onnx.helper.set_model_props(model_proto, metadata)
+
+                _save_protobuf(export_path, model_proto)
+
+                t1 = time.time()
+                logging.info(f"Exported onnx in {round((t1-t0)*1000)} ms")
```
singhblom, line 203: Since this is a callback maybe use logging callback instead here?

emote/nn/gaussian_policy.py
```diff
@@ -61,22 +64,25 @@ def forward(self, x: Tensor) -> Tuple[Tensor, float]:
...
-        assert log_prob.shape == (bsz, 1)
+            return sample, log_prob
 
-        return sample, log_prob
+        return torch.tanh(mean + std * epsilon)
 
 
 class GaussianMlpPolicy(nn.Module):
-    def __init__(self, observation_dim, action_dim, hidden_dims):
+    def __init__(self, observation_dim: int, action_dim: int, hidden_dims: list[int]):
```
singhblom, line 85: Python 3.8 does not support the `list` type syntax, has to be `List` until we drop 3.8.

emote/proxies.py
```diff
@@ -17,6 +21,14 @@ def __call__(
         """Take observations for the active agents and returns the relevant network output."""
         ...
 
+    @property
+    def policy(self) -> nn.Module:
+        pass
+
+    @property
+    def input_names(self) -> tuple[str]:
```
singhblom, line 29: 3.8 vs 3.9

tests/test_htm.py
```diff
@@ -33,22 +36,42 @@ def forward(self, action, obs):
...
+def test_htm():
+    device = torch.device("cpu")
+    env = DictGymWrapper(AsyncVectorEnv(10 * [HitTheMiddle]))
+    table = DictObsTable(spaces=env.dict_space, maxlen=1000, device=device)
+    memory_proxy = TableMemoryProxy(table)
+    dataloader = MemoryLoader(table, 100, 2, "batch_size")
+
+    q1 = QNet(2, 1)
+    q2 = QNet(2, 1)
+    policy = Policy(2, 1, [N_HIDDEN, N_HIDDEN])
```
singhblom, line 48: I think I was trying to get away from having the "smart" convenience class `GaussianMLlpPolicy`. I think it's a mistake to have those kinds of convenience wrappers, it is much better to make constructing a policy very easy. But but, it's still there. I can refactor it later.

tests/test_htm.py
```diff
@@ -57,10 +80,18 @@ def test_htm():
 
     q1 = QNet(2, 1)
     q2 = QNet(2, 1)
-    policy = Policy(2, 1)
+    policy = Policy(2, 1, [N_HIDDEN, N_HIDDEN])
     ln_alpha = torch.tensor(1.0, requires_grad=True)
     agent_proxy = FeatureAgentProxy(policy, device)
 
+    exporter = OnnxExporter(
```
singhblom, line 87: Noice!

singhblom (approved):
Excellent job. I think the type hints need to be fixed but apart from that feel free to think about the comments or not, whatever you feel like.

tests/test_htm.py
tgolsson: Yepp, let's think about that later.

PR #81, opened 2023-02-17
add performance logger

tgolsson:
A plugin for logging the RAM, CPU and (later GPU) usage of the process.


emote/extras/memory_logger.py
```diff
@@ -0,0 +1,29 @@
...
+class MemoryLogger(LoggingCallback):
+    def __init__(self):
+        super().__init__(cycle=1_000)
+        self._proc = psutil.Process(os.getpid())
+        self._previous_memory = self._proc.memory_info().rss / (1024 * 1024)
+        self._previous_bp_step = 0
+
+    def on_cycle_end(self, bp_step, bp_samples):
+        memory_now = self._proc.memory_info().rss / (1024 * 1024)
+        self.log_scalar("ram/usage_mb", memory_now)
```
singhblom, line 21: Why do you have different namespaces for these two logs? Put both under "memory_stats" in my opinion.

tgolsson: Oh I missed one. I think I'll do `system/ram`. We have traditionally had `memory` for the replay memory, so it makes it confusing to have `memory_stats` which is why I switched it away from that are.

singhblom (approved):

PR #82, opened 2023-02-21
changed test to check reward instead of loss

spetravic:
Created a FinalRewardTestCheck callback that checks the scaled reward. The FinalLossTestCheck in the HTM test is replaced with FinalRewardTestCheck which raises an exception if the reward is lower than -5.


singhblom (approved):
Nice!

tgolsson (approved):
LGTM! Thank you :-)

PR #87, opened 2023-03-01
Change minimal version to Python3.9

tgolsson:


jaxs-ribs (approved):

emote/proxies.py
```diff
@@ -2,7 +2,6 @@
 Proxies are bridges between the world the agent acts in and the algorithm training loop.
 """
 
-from __future__ import annotations
 
 from typing import Dict, Protocol
```
singhblom, line 6: You don't want to update the type annotations? `Dict -> dict` etc.

tgolsson: I was considering doing all files but it seems like something we can do incrementally.

PR #90, opened 2023-03-03
use input_key in FeatureAgentProxy.input_names

tgolsson:
Missed connection between two of my overlapping PRs

PR #91, opened 2023-03-03
allow None in cycle to disable

tgolsson:
I think we talked about what cycle 0 means before @singhblom but I added a bug in the OnnxExporter by default to None instead of 0. And I think that is *more* correct as a value for "don't run".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bugfixes
 
 - Fix `FeatureAgentProxy.input_names` to use `input_key` when configured.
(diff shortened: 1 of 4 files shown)


tgolsson:
Current dependencies on/for this PR:
* main
  * **PR #90** <a href="https://app.graphite.dev/github/pr/EmbarkStudios/emote/90" target="_blank">Graphite</a> 
    * **PR #91** <a href="https://app.graphite.dev/github/pr/EmbarkStudios/emote/91" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/EmbarkStudios/emote/91?utm_source=stack-comment).

singhblom (approved):

PR #93, opened 2023-03-08
refactor LoggingCallback to LoggingMixin

tgolsson:
This changes LoggingCallback to be a mixin class instead. This decouples it from the callback system, making it reusable as a baseclass for proxies and other objects that live on the inference side.

### DIFF
--- a/emote/callbacks.py
+++ b/emote/callbacks.py
@@ -15,13 +15,14 @@ from .callback import Callback
 from .trainer import TrainingShutdownException
 
 
-class LoggingCallback(Callback):
-    """A Callback that accepts logging calls.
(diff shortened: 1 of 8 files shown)


emote/callbacks.py
```diff
@@ -15,13 +15,14 @@
 from .trainer import TrainingShutdownException
 
 
-class LoggingCallback(Callback):
-    """A Callback that accepts logging calls.
+class LoggingMixin:
```
singhblom, line 18: I don't think this should be in `callbacks.py` when it's no longer a callback.

singhblom (approved):
Nice!

PR #96, opened 2023-03-09
add LoggingProxyWrapper for the memory

tgolsson:
Also adds a short document about metrics to the docs. :-)


jaxs-ribs (approved):

emote/callbacks.py
```diff
@@ -24,20 +24,44 @@ class LoggingMixin:
...
+
     def log_scalar(self, key: str, value: Union[float, torch.Tensor]):
         """Use log_scalar to periodically log scalar data."""
         if isinstance(value, torch.Tensor):
             self.scalar_logs[key] = value.item()
         else:
             self.scalar_logs[key] = value
 
+    def log_windowed_scalar(self, key: str, value: Union[float, torch.Tensor]):
+        """Use log_windowed_scalar to periodically log scalar data."""
```
singhblom, line 47: Maybe we could mention that this actually provides a time series. With this docstring it's not clear what the difference is between the windowed_scalar and the scalar version.

singhblom (approved):
lgtm.

emote/callbacks.py
tgolsson: Next PR..! Just kidding. Will fix :-)

singhblom (approved):
Neato!

PR #101, opened 2023-03-10
Ali/mbrl lib

Ali-Ghadirzadeh:
The implementation of the model-based RL code.


emote/models/ensemble.py
```diff
@@ -0,0 +1,351 @@
...
+        *args,
+        **kwargs,
+    ):
+        super().__init__()
+        self.in_size = in_size
+        self.out_size = out_size
+        self.num_members = num_members
+        self.propagation_method = propagation_method
+        self.propagation_indices: torch.Tensor = None
+        self.elite_models: list[int] = None
```
tgolsson, line 118: Lies. It's not a list.

emote/models/ensemble.py
```diff
@@ -0,0 +1,351 @@
...
+            mean, logvar = self._default_forward(x, only_elite=False)
+            if self.num_members == 1:
+                mean = mean[0]
+                logvar = logvar[0] if logvar is not None else None
+            return mean, logvar
+        assert x.ndim == 2
+        model_len = (
+            len(self.elite_models) if self.elite_models is not None else len(self)
+        )
+        if x.shape[0] % model_len != 0:
```
tgolsson, line 297: Maybe should be assert? :-)

emote/models/ensemble.py
```diff
@@ -0,0 +1,351 @@
...
+                f"{model_len} models."
+            )
+        x = x.unsqueeze(0)
+        if self.propagation_method == "random_model":
+            # passing generator causes segmentation fault
+            # see https://github.com/pytorch/pytorch/issues/44714
+            model_indices = torch.randperm(x.shape[1], device=self.device)
+            return self._forward_from_indices(x, model_indices)
+        if self.propagation_method == "fixed_model":
+            if self.propagation_indices is None:
```
tgolsson, line 310: Can be checked in init?

emote/models/ensemble.py
```diff
@@ -0,0 +1,351 @@
...
+                raise ValueError(
+                    "When using propagation='fixed_model', `propagation_indices` must be provided."
+                )
+            return self._forward_from_indices(x, self.propagation_indices)
+        if self.propagation_method == "expectation":
+            mean, logvar = self._default_forward(x, only_elite=True)
+            return mean.mean(dim=0), logvar.mean(dim=0)
+        raise ValueError(f"Invalid propagation method {self.propagation_method}.")
```
tgolsson, line 318: Can be checked in init.

tgolsson (changes requested):
OK; a lot to unpack! Looks good, with some concerns about datagroups etc we'll talk about next week!

I also did a quick scan on a more nitpick level. In general, the code feels quite dense with lots of branching, which makes it harder to understand, debug and maintain for others. If we can cut out some branches we don't use, some arguments we don't need, and add some straighter left margins I think it'd help a ton with readability of it all. We can think about this together if you want :-)

I also believe having a "simple" example would help a lot with understanding the higher level flow! I think I see the bigger puzzle pieces and flow, but I'm not quite seeing the ship yet even with the Lunar Lander integration. The comments definitely help there though. And tests please. 🙏 

Great work so far! Have a nice weekend :-)

tgolsson:
Fixed some tests, if you rebase or merge main those should get resolved. :-)

emote/callback.py
```diff
@@ -78,13 +78,12 @@ def _wrap_callback_function(obj, func, *, group: str = None, use_group: bool = T
     # backward needs to pass things to loss so treated specially.
     # TODO(singhblom) Figure out if this is the nicest way to do it.
     if args.varargs or args.varkw:
-        if func.__name__ != "backward":
+        if func.__name__ != "backward" and func.__name__ != "begin_batch":
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: I'm not sure why we need to raise deprecation warning here. Can't we remove the warning?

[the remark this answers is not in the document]
tgolsson: No, *args and **kwargs is an antipattern that makes debuggability worse :-) We should optimally have no escape hatch for backward too; it's just too noisy historically.

https://github.com/EmbarkStudios/cog/pull/264 for the original PR in cog.

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+        Args:
+            action (tensor): the action at.
+            observation (tensor): the observation/state st.
+            rng (torch.Generator): a random number generator.
+        Returns:
+            (tuple): predicted observation and rewards.
+        """
+        obs = to_tensor(observation).to(self.device)
+        model_in = self.get_model_input(obs, action)
+        if not hasattr(self.model, "sample"):
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed!

emote/callback.py
```diff
@@ -78,13 +78,12 @@ def _wrap_callback_function(obj, func, *, group: str = None, use_group: bool = T
...
             warnings.warn(
                 f"Deprecated: {func.__qualname__} uses *args or **kwargs, this is deprecated",
                 UserWarning,
             )
         return func
-
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: Fixed

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+        self,
+        model: nn.Module,
+        target_is_delta: bool = True,
+        normalize: bool = False,
+        normalize_double_precision: bool = False,
+        learned_rewards: bool = True,
+        obs_process_fn: Optional[nn.Module] = None,
+        no_delta_list: Optional[list[int]] = None,
+    ):
+        super(DynamicModel, self).__init__()
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+        self.obs_process_fn = obs_process_fn
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> tuple[torch.Tensor, ...]:
+        """Computes the output of the dynamics model.
+        Args:
+            x (tensor): input
+        Returns:
+            (tuple of tensors): predicted tensors
+        """
+        return self.model.forward(x, *args, **kwargs)
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: Fixed
Removed args, kwargs

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+            (tuple): predicted observation and rewards.
+        """
+        obs = to_tensor(observation).to(self.device)
+        model_in = self.get_model_input(obs, action)
+        if not hasattr(self.model, "sample"):
+            raise RuntimeError(
+                "DynamicModel requires wrapped model to implement sample method"
+            )
+        preds = self.model.sample(model_in, rng)
+        if len(preds.shape) != 2:
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: Fixed

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+            raise RuntimeError(
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed!

emote/models/model_env.py
```diff
@@ -0,0 +1,315 @@
...
+            truncated_linear(*(self._rollout_schedule + [self.bp_counter + 1]))
+        )
+        if self._len_rollout != len_rollout:
+            self._len_rollout = len_rollout
+            new_memory_size = (
+                self._len_rollout
+                * self._model_env.num_envs
+                * self.num_bp_to_retain_buffer
+            )
+            self._memory.resize(new_memory_size)
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: The general idea is to overwrite the entire memory after a **fixed number of BP steps**.  However, by gradually increasing the rollout-size, we generate more transitions of (s_t, a_t, s_{t+1}) in every BP step, and as the result we overwrite the memory faster. By resizing the memory, it is guaranteed that samples are not overwritten before a given number of BP steps passed.

emote/typing.py
```diff
@@ -1,10 +1,18 @@
...
-from typing import Dict, List
+from typing import Callable, Dict, List, Union
+
+import numpy as np
+import torch
 
 from numpy.typing import ArrayLike
 
 
+TensorType = Union[torch.Tensor, np.ndarray]
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: I'll try to make it more restrict.

emote/models/model_env.py
```diff
@@ -0,0 +1,315 @@
...
+from typing import Optional, Union
+
+import gymnasium as gym
+
+# This file contains codes and texts that are copied from
+# https://github.com/facebookresearch/mbrl-lib
+import numpy as np
+import torch
+
+from tests.gym.collector import CollectorCallback
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed

emote/models/ensemble.py
```diff
@@ -0,0 +1,351 @@
...
+            xw = x.matmul(self.weight)
+            if self.bias is not None:
+                return xw + self.bias
+            else:
+                return xw
+
+    def set_elite(self, elite_models: Sequence[int]):
+        self.elite_models = list(elite_models)
+
+    def toggle_use_only_elite(self):
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: elite models are completely removed. 
We are not using them at this point.

emote/models/ensemble.py
Ali-Ghadirzadeh: fixed

emote/models/ensemble.py
Ali-Ghadirzadeh: We will only use expectation propagation, and all other methods are removed.
Fixed then!

emote/models/ensemble.py
Ali-Ghadirzadeh: Fixed! Actually it was not needed at all.

emote/utils/math.py
```diff
@@ -0,0 +1,143 @@
...
+            stats = pickle.load(f)
+            self.mean = torch.from_numpy(stats["mean"]).to(self.device)
+            self.std = torch.from_numpy(stats["std"]).to(self.device)
+
+    def save(self, save_dir: str):
+        """Saves stored statistics to the given path."""
+        with open(save_dir, "wb") as f:
+            pickle.dump(
+                {"mean": self.mean.cpu().numpy(), "std": self.std.cpu().numpy()}, f
+            )
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed

emote/models/ensemble.py
```diff
@@ -0,0 +1,351 @@
...
+            if self.bias is not None:
+                return xw + self.bias[self.elite_models, ...]
+            else:
+                return xw
+        else:
+            xw = x.matmul(self.weight)
+            if self.bias is not None:
+                return xw + self.bias
+            else:
+                return xw
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed

emote/callback.py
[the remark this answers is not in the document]
Ali-Ghadirzadeh: I removed the added part. Now, we receive warnings about this. As we discussed, we will fix it by another PR.

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+                    preds.shape
+                )
+            )
+        next_observs = preds[:, :-1] if self.learned_rewards else preds
+        if self.target_is_delta:
+            tmp_ = next_observs + obs
+            for dim in self.no_delta_list:
+                tmp_[:, dim] = next_observs[:, dim]
+            next_observs = tmp_
+        rewards = preds[:, -1:] if self.learned_rewards else None
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: The main idea of having a wrapper class is for the model not to care about state, actions and rewards. The model simply is a regression model that maps some inputs to some outputs :) So delegating knowing reward to model is not inline with this approach. 
To simplify the code, I removed the "target_is_delta" flag, and assume that target is always delta. Then, we'll only have no_delta_mask for the dimensions of the obs that should not be delta.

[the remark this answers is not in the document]
Ali-Ghadirzadeh: masking is a bit more complicated than a list of dimension because of the batch size.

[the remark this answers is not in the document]
tgolsson: Alrighty, makes sense! I figured with broadcasting it might work out fine (maybe bmm...) but if you think this is better/easier/clearer I'll defer to you :-)

emote/models/model.py
```diff
@@ -0,0 +1,241 @@
...
+        obs_process_fn (callable, optional): if provided, observations will be passed through
+            this function before being given to the model.
+        no_delta_list (list(int), optional): if provided, represents a list of dimensions over
+            which the model predicts the actual observation and not just a delta.
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        target_is_delta: bool = True,
```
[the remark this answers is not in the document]
Ali-Ghadirzadeh: fixed

emote/models/model.py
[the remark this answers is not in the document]
Ali-Ghadirzadeh: I resolve it for now to be able to continue. Please feel free to bring it up again if needed

emote/memory/table.py
```diff
@@ -70,6 +70,11 @@ def __init__(
 
         self.clear()
 
+    def resize(self, new_size):
+        with self._lock:
+            if new_size > self._maxlen:
+                self._maxlen = new_size
```
tgolsson, line 76: I think this should either raise a value error or shrink the memory if it's reduced. Not just silently ignore it.

experiments/gym/train_lunar_lander.py
```diff
@@ -80,108 +86,241 @@ def forward(self, obs):
...
+        Arguments:
+            args:
+            train_cbs (list[Callback]):
+        Returns:
+            (list[Callback]): the full list of callbacks for the training
+    """
+    callbacks = train_cbs + [
         TensorboardLogger(
-            logged_cbs,
+            train_cbs,
```
tgolsson, line 251: This seems wrong. The logger should only take logger cbs (things that inherit `LoggingMixin`)

emote/utils/model.py
```diff
@@ -0,0 +1,24 @@
...
+        return torch.from_numpy(x)
+    raise ValueError("Input must be torch.Tensor or np.ndarray.")
+
+
+def to_numpy(x: Union[torch.Tensor, np.ndarray]):
+    if isinstance(x, torch.Tensor):
+        return x.detach().to("cpu").numpy()
+    if isinstance(x, np.ndarray):
+        return x
+    raise ValueError("Input must be torch.Tensor or np.ndarray.")
```
tgolsson, line 24: Are we still needing these?

emote/typing.py
```diff
@@ -47,6 +52,20 @@ class EpisodeState(Enum):
...
+        self,
+        bp_step_begin: float,
+        bp_step_end: float,
+        value_min: float,
+        value_max: float,
+    ):
+        self.bp_step_begin = bp_step_begin
+        self.bp_step_end = bp_step_end
+        self.value_min = value_min
+        self.value_max = value_max
```
tgolsson, line 66: Use a dataclass please. :-) Also, it *feels* like (a) this should be in something like `emote.extra.schedules`, and (b) it should have a evaluate_at or similar function? :-)

emote/sac.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         roll_length: int = 1,
     ):
         super().__init__()
+        self._order = 1  # this is to ensure that the data_group is prepared beforehand
```
tgolsson, line 105: Because we have a collector callback, right? Where does the issue occur otherwise?

emote/models/model_env.py
```diff
@@ -0,0 +1,197 @@
...
+        batched_actions = np.stack(
+            [actions[agent].list_data["actions"] for agent in self._agent_ids]
+        )
+        next_obs, rewards, dones, info = self.step(batched_actions)
+        new_agents = []
+        results = {}
+        terminated = info["terminated"]
+
+        for env_id, (done, term) in enumerate(zip(dones, terminated)):
+            if done or term:
```
tgolsson, line 143: Nit: I'd not use terminated/term because below it reads like a bug that term isn't linked with terminal. Human pattern-seeking and all that. I'd at least add a comment that `terminated` means the time ran out, not that it hit a terminal state.

emote/models/model.py
```diff
@@ -0,0 +1,223 @@
...
+            optimizer=opt,
+            lr_schedule=lr_schedule,
+            network=model,
+            max_grad_norm=max_grad_norm,
+            data_group=data_group,
+        )
+        self.model = model
+
+    def loss(self, observation, next_observation, actions, rewards):
+        loss, loss_info = self.model.loss(
```
tgolsson, line 217: Nit: use `_` to make it clear you're intentionally ignoring the second result.

emote/utils/model.py
```diff
@@ -0,0 +1,24 @@
...
+    if isinstance(x, torch.Tensor):
+        return x
+    if isinstance(x, np.ndarray):
+        return torch.from_numpy(x)
+    raise ValueError("Input must be torch.Tensor or np.ndarray.")
+
+
+def to_numpy(x: Union[torch.Tensor, np.ndarray]):
+    if isinstance(x, torch.Tensor):
+        return x.detach().to("cpu").numpy()
```
tgolsson, line 21: If we still need them, do we actually need to explicitly move to CPU before going to numpy?

tgolsson (changes requested):
Much much improved! Awesome work 🥳 I think you're missing a __init__.py in emote.models to ensure it gets picked up by all tools, and there's some small nits and questions left still.

emote/sac.py
tgolsson: Wondering if we can fix this in the trainer somehow or by making it more closely tied with the loader.

experiments/gym/train_lunar_lander.py
Ali-Ghadirzadeh: Sorry didn't get what is wrong here. 
All callbacks in train_cbs (LossCallback, BatchCallback, GymCollector) inherit LoggingMixin and Callback.

emote/utils/model.py
Ali-Ghadirzadeh: Yes, I think so. Here is an example:
b = a.detach().numpy()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.

emote/utils/model.py
Ali-Ghadirzadeh: It is still nice to have a function for 'to_numpy()'. However, since it is a single line code, I can also remove the function. This is how it looks like now:
def to_numpy(x: torch.Tensor):
    return x.detach().to("cpu").numpy()

emote/typing.py
Ali-Ghadirzadeh: Thanks for the great suggestion! Fixed it :)

emote/utils/model.py
tgolsson: Ok, sounds good then!

emote/utils/model.py
tgolsson: Hmm. We can keep it for now and see how it feels. I'm generally more for explicit code, but it's quite verbose.

emote/sac.py
Ali-Ghadirzadeh: The QTarget needs the RL batch at the begin_batch. However, the RL batch is also created at the begin_batch by the BatchSampler. So, an error will occur in case QTarget.begin_batch is called before BatchSampler.begin_batch. So, I set the _order=1 for QTarget to ensure BatchSampler is always called first (its _order is zero by default).

emote/models/model_env.py
Ali-Ghadirzadeh: Fair point :)
I renamed it to reached_max_len

emote/sac.py
tgolsson: Righto! I'll think about this, but we can keep it like it is for now.

tgolsson (approved):
LGTM! Nice :-)

Ali-Ghadirzadeh:
Merging MBRL implementation into main

PR #104, opened 2023-03-13
Weights&biases logging integration

spetravic:
Created a new callback WBLogger which is based on the TensorboardLogger and logs the same things to Weights&Biases.

In order to not make _wandb_ mandatory as a dependency in case someone doesn't want to use it at all, it is only imported if the flag `--use_wandb` is set to True. As imports are always supposed to be at the top of the file, I am not sure if this is the proper way to do it, so suggestions are welcome.

Usage notes:
After installing wandb, you can login by running `wandb login` in the terminal and pasting the API key found in your W&B user settings.

Note that the entity to be used for the wandb initialization (where to create the runs/projects) is not hardcoded. Instead, it is taken as the default setting of the user, which should be set to _embark_ for the work accounts by default. If needed it can be changed either in the W&B user settings or by passing `entity="embark"` to `wandb.init()`.

### DIFF
--- a/.buildkite/validate-lockfile.sh
+++ b/.buildkite/validate-lockfile.sh
@@ -3,42 +3,17 @@ set -eo pipefail
 pdm plugin add pdm-plugin-torch==23.0
 
 EXIT_CODE=0
-pdm lock --check || EXIT_CODE=$?
-
-if [ $EXIT_CODE -ne 0 ]; then
-    buildkite-agent annotate --style "error" --context "lockfile" ":lock: Failed validating lockfile. See logs for more info."
-	exit 1
-fi
-
 TORCH_EXIT_CODE=0
+
+pdm lock --check || EXIT_CODE=$?
 pdm torch lock --check || TORCH_EXIT_CODE=$?
-GIT_STATUS=$(git status --porcelain --untracked-files=no -- pdm.lock)
-if [ -n "$GIT_STATUS" ] && [ $TORCH_EXIT_CODE -ne 0 ]; then
-	lock_diff=$(git diff pdm.lock)
-	cat << EOF | buildkite-agent annotate --style "error" --context "lockfile"
-:lock: Lockfiles are outdated. Please run \`pdm lock --no-update && pdm torch lock\` and commit the result.
 
-\`\`\`diff
-$lock_diff
-\`\`\`
-EOF
-	exit 1
-elif  [ -n "$GIT_STATUS" ] ; then
-	lock_diff=$(git diff pdm.lock)
+if [ $EXIT_CODE -ne 0 ] || [ $TORCH_EXIT_CODE -ne 0 ]; then
 	cat << EOF | buildkite-agent annotate --style "error" --context "lockfile"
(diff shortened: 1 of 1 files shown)


experiments/gym/train_lunar_lander.py
```diff
@@ -163,25 +163,65 @@ def train_lunar_lander(args):
...
+                100,
+            ),
+        ]
+
+    callbacks = logged_cbs + logger + [BackPropStepsTerminator(args.num_bp_steps)]
     trainer = Trainer(callbacks, dataloader)
     trainer.train()
 
+    if args.use_wandb:
+        wandb.finish()
```
tgolsson, line 208: Could this be handled by the Wandb logger?

experiments/gym/train_lunar_lander.py
```diff
@@ -163,25 +163,65 @@ def train_lunar_lander(args):
...
+            "batch_size": batch_size,
+            "learning_rate": learning_rate,
+            "rollout_len": rollout_len,
+        }
+
+        wandb_run = wandb.init(
+            project=config["wandb_project"],
+            name=config["wandb_run"],
+            config=wandb.helper.parse_config(
+                config, exclude=("wandb_project", "wandb_run")
```
tgolsson, line 182: Why are we adding them to the config 10 lines up and then excluding them here? :D

What does "parse_config" do? It seems weird to give that a python-dict... sounds more like it would want a string?

experiments/gym/train_lunar_lander.py
```diff
@@ -163,25 +163,65 @@ def train_lunar_lander(args):
...
+        )
 
+        logger = [
+            WBLogger(
+                callbacks=logged_cbs,
+                log_interval=100,
+            ),
+        ]
+    else:
+        logger = [
```
tgolsson, line 193: Nit: no need for a list here; you can just add it to the list after BackPropStepsTerminator instead. A bit less spacing.

experiments/gym/train_lunar_lander.py
spetravic: I was sending the full config to the logger first, when I switched it I must have forgotten to rewrite this part [🤦](https://emojipedia.org/person-facepalming/) 

Parse_config is used to exclude some keys from the dictionary so they are not logged as separate fields in the W&B table. It used to be a parameter passed to init(), but it is deprecated and they suggest using this instead.

tgolsson: Nice! Do you know how it handles recursiveness? :-) I.e, if we have a dict inside the config dict.

spetravic: The way it works is the following: `config = {key: value for key, value in config.items() if key not in exclude}`, so in the case of a dict inside config, we could exclude the entire dict based on its key, but not a key inside of it.

tgolsson: 👍 Thank you! Sounds great.

tgolsson (approved):
LGTM! Nice work 🎉

PR #105, opened 2023-03-15
simplify lockfile error

tgolsson:
With `pdm lock --check` setting since #62 and #76 automating the pdm torch lock this wasn't giving the right or a helpful error message.


spetravic (approved):
Great, thank you!

jaxs-ribs (approved):

PR #107, opened 2023-03-17
Improve CI error messages

tgolsson:


.buildkite/run-isort.sh
```diff
@@ -10,7 +10,7 @@ cat diff.txt
 
 if [ $EXIT_CODE -ne 0 ]; then
 	cat << EOF | buildkite-agent annotate --style "error" --context "isort"
-:warning: Your imports aren't sorted  by \`isort\`. Please fix the below diffs, or run \`pdm run isort emote\` to automatically format it.
+:warning: Your imports aren't sorted  by \`isort\`. Please fix the below diffs, or run \`pdm run isort emote experiments\` to automatically format it.
```
Ali-Ghadirzadeh, line 13: do we need to run isort on tests\ as well? 
This is what the docs says:
pdm run isort emote/ tests/

tgolsson: Good catch, added! Thank you.

singhblom (approved):

PR #109, opened 2023-03-24
split callbacks.py into a package

tgolsson:
The callbacks.py file was getting a bit too large so this breaks it down into a package with dedicated files per domain.

Also adds some more docs and probably picks up some lints.

Also makes the import error for wandb a bit more helpful by chaining it with some info in case it's not imported and adds a test case for it.


singhblom:
Are we still waiting for my review here? Should we get back to this?

tgolsson:
Just haven't had time to revisit!

tgolsson:
@singhblom Updated and fixed your comments!

emote/callbacks/logging.py
```diff
@@ -0,0 +1,107 @@
...
+from emote.callback import Callback
+from emote.mixins.logging import LoggingMixin
+
+
+class TensorboardLogger(Callback):
+    """Logs the provided loggable callbacks to tensorboard."""
+
+    def __init__(
+        self,
+        callbacks: List[LoggingMixin],
```
singhblom, line 17: Call this `loggables` or something?

emote/callbacks/logging.py
```diff
@@ -0,0 +1,107 @@
...
+        writer: SummaryWriter,
+        log_interval: int,
+        log_by_samples: bool = False,
+    ):
+        super().__init__(cycle=log_interval)
+        self._cbs = callbacks
```
singhblom, line 23: And then say `self._logs = loggables` like in the terminal logger.

singhblom (approved):
lgtm but two minor nitpicks.

PR #110, opened 2023-03-27
Ali/mbrl tests

ghost:
Test functions for the MBRL implementation


emote/callbacks.py
```diff
@@ -502,3 +502,19 @@ def end_cycle(self):
...
+
+class BatchCallback(LoggingMixin, Callback):
+    def __init__(self):
+        super().__init__()
+
+    def begin_batch(self, *args, **kwargs):
+        pass
+
+    @Callback.extend
+    def collect_multiple(self, *args, **kwargs):
```
tgolsson, line 515: This feels a bit like an implementation detail? Or do we expect collecting to be an important part of the API, as opposed to getting batches?

emote/models/model.py
```diff
@@ -175,43 +187,143 @@ def load(self, load_dir: str) -> None:
...
+                    nn.Linear(hidden_size, hidden_size),
+                    nn.BatchNorm1d(hidden_size),
+                    nn.ReLU(),
+                )
+            )
+        network.append(nn.Sequential(nn.Linear(hidden_size, out_size)))
+        self.network = nn.Sequential(*network).to(self.device)
+        self.init_param()
+
+    def init_param(self):
```
tgolsson, line 223: This seems like an anti-pattern, why do you need a separate function? It's literally below.

tgolsson (commented):
Did a quick scan, just some minor notes - I think they're all from the previous PR; just noticed now. Will do a deeper check of the tests tomorrow.

tests/test_gym_collector.py
```diff
@@ -0,0 +1,109 @@
...
+            obs_err = torch.mean(torch.abs(sim_next_obs[i] - next_obs[i])).detach()
+            reward_err = torch.mean(torch.abs(sim_reward[i] - reward[i])).detach()
+            if obs_err > 0.001 or reward_err > 0.001:
+                print(f"obs_err: {obs_err}, reward_err: {reward_err}")
+                print(f"obs: {obs[i]}, action: {action[i]}")
+                print(f"reward: {reward[i]}\nsim_reward: {sim_reward[i]}")
+                print(f"next_obs: {next_obs[i]}\nsim_next_obs: {sim_next_obs[i]}")
+                raise ValueError(
+                    "Loaded values for obs/reward does not match the calculated ones"
+                )
```
tgolsson, line 52: I'd suggest doing this on the following format:

```
message = f'''
    obs_err ..
    obs ...
    reward ...
'''
raise ValueError("Loaded... {message}")
```

That will make it clearer in pytest I think, in case there's lots of logging etc. In py3.11 this is an *excellent* use-case for PEP678 too.

tests/test_gym_collector.py
```diff
@@ -0,0 +1,109 @@
...
+        raise TrainingShutdownException()
+
+    def get_batch(self, observation, next_observation, actions, rewards):
+        return observation["obs"], next_observation["obs"], actions, rewards
+
+
+def test_gym_collector():
+    device = torch.device("cpu")
+    batch_size = 5
+    rollout_length = 1
```
tgolsson, line 84: I'd add a note that the code doesn't work when this is changed.

tests/test_mbrl.py
```diff
@@ -0,0 +1,206 @@
...
+    The test checks the following:
+        * the replay buffer contains a correct number of samples,
+        * stored samples are the ones generated by the fake model.
+
+    """
+    num_obs = 2
+    num_actions = 1
+    batch_size = 10
+    rollout_size = 5
+    rl_data_group = "rl_group"
```
tgolsson, line 111: Are these constants or config? I'd consider moving them out of the function and capitalize them to denote them as "constants".

tgolsson (changes requested):
Some minor nits, but overall looks good (with the above notes too).

emote/models/ensemble.py
```diff
@@ -84,7 +82,7 @@ def __init__(
         device: Union[str, torch.device],
         num_layers: int = 4,
         ensemble_size: int = 1,
-        hid_size: int = 200,
+        hid_size: int = 256,
```
singhblom, line 85: I prefer if these kinds of things are written out, so `hidden_size`, as long as it is not a very standard acronym.

singhblom (approved):
Nice! Clean code, looks great. One miner nitpick that you may deal with if you want to.

emote/callbacks.py
```diff
@@ -502,3 +502,19 @@ def end_cycle(self):
         if reward < self._cutoff:
             raise Exception(f"Reward too low: {reward}")
         raise TrainingShutdownException()
+
+
+class BatchCallback(LoggingMixin, Callback):
+    def __init__(self):
+        super().__init__()
+
+    def begin_batch(self, *args, **kwargs):
```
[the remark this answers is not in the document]
AliGhadirzadeh: Yes, it is correct. I removed "begin_batch".

emote/callbacks.py
AliGhadirzadeh: You're right! 
Removed "collect_multiple". The callback now only provides the batch via "get_batch"

emote/models/model.py
AliGhadirzadeh: Fair point!
However, it should be also possible to use the init_param function by other NN models. So, I moved the function to utils.model. I also moved truncated_normal_init to utils.model for the same reason.

tests/test_mbrl.py
AliGhadirzadeh: num_obs, num_actions and rl_data_group are constants and are moved outside functions. However, batch_size, and rollout_size have different values for the two functions. So, they are kept inside the functions.

emote/models/ensemble.py
AliGhadirzadeh: fixed!

tests/test_gym_collector.py
AliGhadirzadeh: Fixed!

tgolsson (approved):
LGTM!

PR #111, opened 2023-04-04
Missing return statement in AlphaLoss

spetravic:
When trying to load a state dictionary for the alpha loss (e.g. from checkpoints), there was a problem with a None value being assigned. This came from a missing return statement in the function where the state dictionary is loaded, which is added here.


tgolsson (approved):
Thanks!

PR #114, opened 2023-04-18
fixed the input_key error

AliGhadirzadeh:
Currently, input_key is hard-coded to "obs" for all mbrl functionalities. The new changes enable choosing input_keys as an extra arguments to the corresponding classes.


tgolsson (approved):
LGTM!

PR #117, opened 2023-04-28
updating the checkpointers

AliGhadirzadeh:
Currently, it is not possible to save/load checkpoints of a single (or a number of) neural network model(s). This feature is needed for finetuning/offline training/genrl in which we only need to load some of the networks and not all of them. This PR enables checkpointing a given network while keeping the possibility of checkpointing everything inside callbacks (as before).


singhblom (approved):
Yes, having the networks as a dict seems better!

tgolsson (changes requested):
Going to hard disagree on what @singhblom said! I'd prefer if we modeled this like Logging Callbacks and the Loggers, i.e., we take a list of objects that know how to get checkpointed. One abstraction model.

singhblom:
Yes, that's fair enough. One reason I like dicts here is that they are a bit self documenting. For a list you basically have to look in the code to see what the order of the networks is.

tgolsson:
You can use cb.name still, like it's done in the tests.

AliGhadirzadeh:
Made following changes to checkpointer and checkpoitloader: 
1. forced the checkpointer to checkpoint only callbacks with 'name' field. Un-named callbacks are ignored with a warning message. 
2. removed the list of names as arguments. Currently, only the name of callback is used for checkpointing and loading.

emote/callbacks.py
```diff
@@ -424,25 +431,43 @@ def __init__(
...
-            net.load_state_dict(state)
-        for opt, state in zip(self._opts, state_dict["optim_state_dicts"]):
-            opt.load_state_dict(state)
+
+        for cb in self._cbs:
+            state = state_dict["callback_state_dicts"][cb.name]
+            if self._only_load_nets and hasattr(cb, "network"):
+                cb.network.load_state_dict(state.pop("network_state_dict"))
+            else:
+                cb.load_state_dict(state)
```
tgolsson, line 469: This else will always load for things that don't have a network. I don't think that's intended. I'd consider passing the flags into the load_state_dict instead. Maybe making it a dataclass with a few fields? The ones we've traditionally used for flagging is the following:

```
@dataclass
class LoadConfiguration:
    """Load the network weights"""

    load_weights: bool = True

    """Load the hparams file containing our steps and other non-ML state"""
    load_hparams: bool = True

    """Load the optimizer state from the checkpoint"""
    load_optimizer_state: bool = True
```

We don't necessarily we have to use the same fields or class, just that these are the three toggles we've traditionally had.

emote/callbacks.py
```diff
@@ -424,25 +431,43 @@ def __init__(
...
+                    f"CheckpointLoader ignored {cb} because of not "
+                    f"having the 'name' field.",
+                    UserWarning,
+                )
+
+        if len(names) != len(set(names)):
+            raise ValueError(
+                "CheckpointLoader is given a list of callbacks with at least"
+                "two callbacks with identical names"
+            )
```
tgolsson, line 459: Very nice!

emote/callbacks.py
tgolsson: I think my concern here is that if I somehow call it _network or have two networks it's going to lead to weird bugs. Having this knowledge in the callback makes more sense.

emote/callbacks.py
```diff
@@ -374,23 +371,37 @@ def __init__(
...
+        for cb in callbacks:
+            if hasattr(cb, "name"):
+                self._cbs.append(cb)
+                names.append(cb.name)
+            else:
+                warnings.warn(
+                    f"Checkpointer ignored {cb} because of not "
+                    f"having the 'name' field.",
+                    UserWarning,
+                )
```
tgolsson, line 391: Wondering whether this should be an exception instead?

Also; what we've done historically is to have 

```
@property
def name(self) -> str:
    return self.__class__.__name__
```

on the callback. So everything has a name by default. That leads to weird issues with duplicated callbacks, so that's why some things have explicit names.

singhblom:
What is the status of this? Should we revive it or close it?

AliGhadirzadeh:
> What is the status of this? Should we revive it or close it?

I'll revive it. It is a bit outdated.

AliGhadirzadeh:
I believe this one is quite outdated and hard to fix. I created a new branch and a new PR for this. I'll delete this one as soon as  the other one is approved,

AliGhadirzadeh:
This is resolved by another one PR. So closing this one.

PR #118, opened 2023-05-04
Hyperbolic tangent instability

spetravic:
The current implementation of the action sampling using _RobustTanhTransform_ causes some numerical instability. This leads to the log probability getting extreme values, which then completely destabilizes the training and the creature fails to learn.

image
image

_TanhTransform_ with cache_size=1 caches the original action values sent to the transform and returns them instead of doing the inverse transformation, which seems to prevent the instability issue. The results are in line with Cog training runs, and the creature is able to learn the tasks.

image
image

### DIFF
--- a/emote/nn/gaussian_policy.py
+++ b/emote/nn/gaussian_policy.py
@@ -13,13 +13,6 @@ from torch import Tensor
 from emote.nn.initialization import ortho_init_, xavier_uniform_init_
 
 
-class RobustTanhTransform(transforms.TanhTransform):
-    def _inverse(self, y):
-        eps = torch.finfo(y.dtype).eps
-        input_val = y.clamp(min=-1.0 + eps, max=1.0 - eps)
-        return torch.atanh(input_val)
-
-
 class BasePolicy(nn.Module):
     def __init__(self):
         super().__init__()
--- a/emote/nn/gaussian_policy.py
+++ b/emote/nn/gaussian_policy.py
@@ -67,7 +67,7 @@ class GaussianPolicyHead(nn.Module):
         if self.training:
             dist = dists.TransformedDistribution(
                 dists.Independent(dists.Normal(mean, std), 1),
-                RobustTanhTransform(),
+                transforms.TanhTransform(cache_size=1),
             )
             sample = dist.rsample()
 


AliGhadirzadeh (approved):
This looks good to me. It should speed up the policy inference as it does not compute tanh and atanh for every single state. Also, gradients are not passed through (a)tanh functions during bp step. Very nice :)

emote/nn/gaussian_policy.py
```diff
@@ -67,7 +67,7 @@ def forward(
         if self.training:
             dist = dists.TransformedDistribution(
                 dists.Independent(dists.Normal(mean, std), 1),
-                RobustTanhTransform(),
+                transforms.TanhTransform(cache_size=1),
```
tgolsson, line 70: You can do this with the RobustTanhTransform too, it derives from this class.

spetravic: I figured since the values are cached and the transformation is not actually inverted, it doesn't make much of a difference (if I am not misunderstanding how the cache works).

singhblom (approved):

emote/nn/gaussian_policy.py
spetravic: I don't know whether to laugh or cry about that, to be honest 🥲

AliGhadirzadeh: I do not think we still need RobustTanhTransform. We do not need to clamp the output of tanh anymore. Even for very large inputs to tanh, the transformation still caches the input and returns the right value for loglikelihood calculations.

tgolsson: What happens if we're running forward and backward passes in parallel though? That's the case for clients which doesn't do local inference during training. In that case the cache would not be valid.

singhblom: Wait, how would the cache not be valid? It's always used on the line after it's set, right?

```python
# Transform created
dist = dists.TransformedDistribution(
    dists.Independent(dists.Normal(mean, std), 1),
    transforms.TanhTransform(cache_size=1),
)
# Transform called, cache set
sample = dist.rsample()
# Inverse called in log_prob, cache used
log_prob = dist.log_prob(sample).view(bsz, 1)
```

singhblom: But I do agree with Ali, if we don't use the RobustTanh we should remove it.

tgolsson: Ah, it's local. Then yeah nvm me. :D Keep thinking like tensorflow with lots of global state.

PR #122, opened 2023-05-10
enforce no logging gradients

tgolsson:
This might be another memory leak that we're missing.


spetravic (approved):

singhblom (commented):
Wait, this I don't like. This feels like a magic function that hides what actually happens in the code. Can we do this check and detach explicitly where it's need, or change things in some way so it's not necessary to detach or something?

tgolsson:
What I did first was just add the asserts, noticed it would be like two dozen places we need to detach; and figured this was the best way to ensure it's painless in the future. I don't think it's necessarily too magic; it's *obvious* you need to detach whenever you store something to log it.

tgolsson:
I realized we already do this in most places because it's a side-effect of `detach`. The only place it doesn't happen is for images. So I'll just do that in a separate PR.

PR #126, opened 2023-05-10
work around memory leak in torch.jit.trace

tgolsson:
Repro'd and confirmed using the methodology here: https://github.com/pytorch/pytorch/issues/82532

We leak every single tensor in the model each iteration. I validated it by inserting the code from the above issue into test_htm_onnx and running `exporter.export` in a loop at the end. This is the objgraph growth per iteration:
```
tuple                16813       +44
dict                 13502       +23
ReferenceType         5569       +22
type                  3021       +22
Tensor                 340        +8
StorageItem             31        +1
StorageItemHandle       31        +1
```
Changing the code in this PR makes the growth only the following:
```
dict                 12921        +1
StorageItem            132        +1
StorageItemHandle      132        +1
```

### DIFF
--- a/emote/extra/onnx_exporter.py
+++ b/emote/extra/onnx_exporter.py
@@ -155,6 +155,10 @@ class OnnxExporter(LoggingMixin, Callback):
 
             self.policy.train(False)
 
+            # NOTE: This might seem like a good use case for torch.jit.trace,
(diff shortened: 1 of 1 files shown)


spetravic (approved):
Nice catch!

PR #128, opened 2023-05-11
Add `checkpointer_index` constructor argument to `Checkpointer`

jaxs-ribs:
There currently isn't the option to pick up from a certain `checkpoint_index` in `Checkpointer`. This can come in useful when wanting to load an existing checkpoint and keep training, adding further checkpoints later on.


tgolsson (approved):
LGTM!

PR #129, opened 2023-05-19
Add support for custom weight decay optimizer

spetravic:
Creates a modified AdamW (that inherits from the standard AdamW) as suggested in _[Efficient deep reinforcement learning requires regulating overfitting](https://arxiv.org/pdf/2304.10466.pdf)_.

ModifiedAdamW disables weight decay for the bias, additional modules (such as LayerNorm), and specific named layers (e.g. the paper suggests excluding the final layer in the Q networks, which can be done by passing `layers_to_exclude = ["final_layer"]`). 

The function for separating the modules was adapted from a [PyTorch minGPT implementation](https://github.com/karpathy/minGPT/blob/3ed14b2cec0dfdad3f4b2831f2b4a86d11aef150/mingpt/model.py#L136) that is MIT licensed.

### DIFF
--- a/emote/optimizers.py
+++ b/emote/optimizers.py
@@ -2,37 +2,44 @@
 # https://github.com/karpathy/minGPT/blob/3ed14b2cec0dfdad3f4b2831f2b4a86d11aef150/mingpt/model.py#L136
 # (MIT license)
 
-from typing import List, Optional, Set, Tuple
+from __future__ import annotations
+
+from typing import Type
 
 import torch
 
 
 def separate_modules_for_weight_decay(
     network: torch.nn.Module,
-    whitelist_weight_modules: Tuple[torch.nn.Module],
-    blacklist_weight_modules: Tuple[torch.nn.Module],
-    layers_to_exclude: Optional[List[str]] = None,
-) -> Tuple[Set, Set]:
+    whitelist_weight_modules: tuple[Type[torch.nn.Module], ...],
+    blacklist_weight_modules: tuple[Type[torch.nn.Module], ...],
+    layers_to_exclude: set[str] | None = None,
+) -> tuple[set[str], set[str]]:
     """Separate the parameters of network into two sets: one set of parameters that will have weight decay, and one set that will not.
 
     Args:
         network (torch.nn.Module): Network whose modules we want to separate.
-        whitelist_weight_modules (Tuple[torch.nn.Module]): Modules that should have weight decay applied to the weights.
-        blacklist_weight_modules (Tuple[torch.nn.Module]): Modules that should not have weight decay applied to the weights.
-        layers_to_exclude (Optional[List[str]], optional): Names of layers that should be excluded. Defaults to None.
+        whitelist_weight_modules (tuple[Type[torch.nn.Module], ...]): Modules that should have weight decay applied to the weights.
+        blacklist_weight_modules (tuple[Type[torch.nn.Module], ...]): Modules that should not have weight decay applied to the weights.
+        layers_to_exclude (set[str] | None, optional): Names of layers that should be excluded. Defaults to None.
 
     Returns:
-        Tuple(Set, Set): Sets of modules with and without weight decay.
+        tuple[set[str], set[str]]: Sets of modules with and without weight decay.
     """
+    # Make sure the same module doesn't appear in both whitelist_weight_modules and blacklist_weight_modules
+    assert (
+        len(set(whitelist_weight_modules) & set(blacklist_weight_modules)) == 0
+    ), "Some modules are both whitelisted and blacklisted!"
 
+    layers_to_exclude = layers_to_exclude or set()
     decay = set()
     no_decay = set()
 
     for mn, m in network.named_modules():
         for pn, p in m.named_parameters():
-            fpn = "%s.%s" % (mn, pn) if mn else pn
+            fpn = f"{mn}.{pn}" if mn else pn
 
-            if layers_to_exclude is not None and mn in layers_to_exclude:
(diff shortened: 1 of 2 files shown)


emote/optimizers.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import List, Optional, Set, Tuple
+
+import torch
+
+
+def separate_modules_for_weight_decay(
+    network: torch.nn.Module,
+    whitelist_weight_modules: Tuple[torch.nn.Module],
+    blacklist_weight_modules: Tuple[torch.nn.Module],
+    layers_to_exclude: Optional[List[str]] = None,
```
tgolsson, line 14: If you add `from __future__ import annotations` you can write this as:
```suggestion
    whitelist_weight_modules: tuple[torch.nn.Module, ...],
    blacklist_weight_modules: tuple[torch.nn.Module, ...],
    layers_to_exclude: set[str] | None = None,
```

which is a bit more modern!

emote/optimizers.py
```diff
@@ -0,0 +1,95 @@
...
+    Returns:
+        Tuple(Set, Set): Sets of modules with and without weight decay.
+    """
+
+    decay = set()
+    no_decay = set()
+
+    for mn, m in network.named_modules():
+        for pn, p in m.named_parameters():
+            fpn = "%s.%s" % (mn, pn) if mn else pn
```
tgolsson, line 33: Do you have to pay per letter? :D

emote/optimizers.py
```diff
@@ -0,0 +1,95 @@
...
+            if layers_to_exclude is not None and mn in layers_to_exclude:
+                # Weights of excluded layers will NOT be weight decayed
+                no_decay.add(fpn)
+            elif pn.endswith("bias"):
+                # Biases will not be decayed
+                no_decay.add(fpn)
+            elif pn.endswith("weight") and isinstance(m, blacklist_weight_modules):
+                # Weights of blacklist modules will NOT be weight decayed
+                no_decay.add(fpn)
+            elif pn.endswith("weight") and isinstance(m, whitelist_weight_modules):
```
tgolsson, line 44: I don't understand the usage of `isinstance` here! Either the type annotation lies or the code is wrong. You're taking a list of objects, but you're checking if m is a type. Should `whitelist_weight_modules` be `tuple[Type[nn.Module]]`?

emote/optimizers.py
```diff
@@ -0,0 +1,95 @@
...
+
+        param_dict = {pn: p for pn, p in network.named_parameters()}
+
+        optim_groups = [
+            {
+                "params": [param_dict[pn] for pn in sorted(list(decay))],
+                "weight_decay": weight_decay,
+            },
+            {
+                "params": [param_dict[pn] for pn in sorted(list(no_decay))],
```
tgolsson, line 90: Is the order important here? It seems to me it wouldn't matter... :-) 

Also, if we need them to be lists here, why not make them lists and sort in the helper function?

tests/test_modified_adamw.py
```diff
@@ -0,0 +1,83 @@
...
+        return self.final_layer(self.encoder(x))
+
+
+def test_module_separation():
+    """Validate that all parameters are added to either "decay" or "no_decay"."""
+    q = QNet(2, 2, [2, 2])
+
+    decay, no_decay = separate_modules_for_weight_decay(
+        q,
+        whitelist_weight_modules=(torch.nn.Linear),
```
tgolsson, line 39: probably missing commas here as well

tests/test_modified_adamw.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            no_wd += 1
+
+    assert (
+        wd == 1
+    ), f"There should be one group that has weight decay, but there are {wd}"
+
+    assert (
+        no_wd == 1
+    ), f"There should be one group that has weight decay, but there are {no_wd}"
```
tgolsson, line 83: Very nice tests! Could you reverse some of these tests and also pass in *bad* data and ensure it fails? There's the pattern

```python
a = {}
with pytest.raises(KeyError):
     a['foo'] # exception gets caught and test passes, but fails the test if the exception isn't caught 
```
You can also ensure the message is what you want using regex, like this:

```python
with pytest.raises(
    ValueError,
    match=r"My custom message: '[a-z]+'"
):
    raise ValueError("My custom message: 'foobar'")    
```

tgolsson (changes requested):
Looks very nice! Super clean, with some nitpicks here and there!

emote/optimizers.py
tgolsson: Also use f-strings!

singhblom (approved):
@tgolsson gave good feedback which I agree with, but apart from that it looks super nice. Looking forward to seeing the results of this. 👍🏼

emote/optimizers.py
```diff
@@ -0,0 +1,95 @@
...
+        layers_to_exclude: List[str] = None,
+    ):
+        decay, no_decay = separate_modules_for_weight_decay(
+            network,
+            whitelist_weight_modules,
+            blacklist_weight_modules,
+            layers_to_exclude,
+        )
+
+        param_dict = {pn: p for pn, p in network.named_parameters()}
```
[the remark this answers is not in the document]
spetravic: `network.named_parameters()` returns a generator, so the optimizer groups would have to be created in a different way. Converting to a dict and using the way it's written now seems cleaner to me than iterating through the modules with `next()` , but I could change it to use the generator instead if that is the preference.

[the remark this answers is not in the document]
spetravic: Yes, that works :)

PR #130, opened 2023-05-22
Add Storage Subdirectory

jaxs-ribs:
This allows optionally specifying the name of the subdirectory for the checkpointer and checkpointloader, with the default being "checkpoints". 

We also create the "checkpoints" subdirectory from here. 

This is done in emote to avoid code duplication in downstream projects and match our existing practices, and because 99% of the time, we want the subfolder to be called "checkpoints".

### DIFF
--- a/emote/callbacks.py
+++ b/emote/callbacks.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import time
 
 from collections import deque
@@ -377,6 +378,7 @@ class Checkpointer(Callback):
         checkpoint_index: int = 0,
         optimizers: Optional[List[optim.Optimizer]] = None,
         networks: Optional[List[nn.Module]] = None,
+        storage_subdirectory: str = "checkpoints"
     ):
         super().__init__(cycle=checkpoint_interval)
         self._cbs = callbacks
@@ -384,6 +386,7 @@ class Checkpointer(Callback):
         self._checkpoint_index = checkpoint_index
         self._opts: List[optim.Optimizer] = optimizers if optimizers else []
         self._nets: List[nn.Module] = networks if networks else []
+        self._storage_subdirectory = storage_subdirectory
 
     def end_cycle(self):
         state_dict = {}
@@ -393,7 +396,9 @@ class Checkpointer(Callback):
         state_dict["training_state"] = {
             "checkpoint_index": self._checkpoint_index,
         }
-        torch.save(state_dict, f"{self._path}.{self._checkpoint_index}.tar")
+        name = f"checkpoint_{self._checkpoint_index}.tar"
+        final_path = os.path.join(self._path, self._storage_subdirectory, name)
+        torch.save(state_dict, final_path)
         self._checkpoint_index += 1
 
 
@@ -427,6 +432,7 @@ class CheckpointLoader(Callback):
         reset_training_steps: bool = False,
         optimizers: Optional[List[optim.Optimizer]] = None,
         networks: Optional[List[nn.Module]] = None,
+        storage_subdirectory: str = "checkpoints"
     ):
         super().__init__()
         self._cbs = callbacks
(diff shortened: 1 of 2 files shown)


emote/callbacks.py
```diff
@@ -377,13 +378,15 @@ def __init__(
         checkpoint_index: int = 0,
         optimizers: Optional[List[optim.Optimizer]] = None,
         networks: Optional[List[nn.Module]] = None,
+        storage_subdirectory: str = "checkpoints"
```
tgolsson, line 381: Should probably rename `path` to `run_root` or similar. And update the docstring above. :-)

emote/callbacks.py
```diff
@@ -393,7 +396,11 @@ def end_cycle(self):
         state_dict["training_state"] = {
             "checkpoint_index": self._checkpoint_index,
         }
-        torch.save(state_dict, f"{self._path}.{self._checkpoint_index}.tar")
+        name = f"checkpoint_{self._checkpoint_index}.tar"
+        folder_path = os.path.join(self._path, self._storage_subdirectory)
+        os.makedirs(folder_path, exist_ok=True)
```
tgolsson, line 401: Do this in begin_training instead.

emote/callbacks.py
```diff
@@ -435,9 +443,14 @@ def __init__(
         self._reset_training_steps = reset_training_steps
         self._opts: List[optim.Optimizer] = optimizers if optimizers else []
         self._nets: List[nn.Module] = networks if networks else []
+        self._stored_subdirectory = storage_subdirectory
 
     def begin_training(self):
-        state_dict: dict = torch.load(f"{self._path}.{self._checkpoint_index}.tar")
+        name = f"checkpoint_{self._checkpoint_index}.tar"
+        folder_path = os.path.join(self._path, self._stored_subdirectory)
+        os.makedirs(folder_path, exist_ok=True)
```
tgolsson, line 451: No need to makedir when loading. But maybe check if it exists. :-)

tgolsson (changes requested):
Some cleanup on docs and some minor nitpicks!

emote/callbacks.py
```diff
@@ -359,7 +359,7 @@ class Checkpointer(Callback):
     callbacks supplied in the constructor.
 
     :param callbacks (List[Callback]): A list of callbacks the should be saved.
-    :param path (str): The path to where the checkpoint should be stored.
+    :param run_root (str): The root path to where the checkpoint should be stored.
```
tgolsson, line 362: You need to add the new param too! And this needs to say where "run artifacts should be stored" or similar.

emote/callbacks.py
```diff
@@ -412,7 +413,7 @@ class CheckpointLoader(Callback):
     probably easier to just do it explicitly when the network is constructed.
 
     :param callbacks (List[Callback]): A list of callbacks the should be restored.
-    :param path (str): The path to where the checkpoint should be stored.
+    :param run_root (str): The root path to where the checkpoint should be stored.
```
tgolsson, line 416: Same here as above. :-)

emote/callbacks.py
```diff
@@ -438,18 +439,17 @@ def __init__(
...
+        self._run_root = run_root
         self._checkpoint_index = checkpoint_index
         self._reset_training_steps = reset_training_steps
         self._opts: List[optim.Optimizer] = optimizers if optimizers else []
         self._nets: List[nn.Module] = networks if networks else []
-        self._stored_subdirectory = storage_subdirectory
+        self._folder_path = os.path.join(run_root, storage_subdirectory)
 
     def begin_training(self):
+        assert(os.path.exists(self._folder_path))
```
tgolsson, line 450: Wrong syntax! I'd also raise an explicit error instead of assert here. :-)

tgolsson:
```python3
class InvalidCheckpointLocation(ValueError):
     pass

raise InvalidCheckpointLocation(f"load location {self._folder_path!r} was specified but does not exist")
```

tests/test_checkpoints.py
```diff
@@ -44,9 +44,9 @@ def onestep_dataloader() -> Generator:
 
 def test_networks_checkpoint():
     chkpt_dir = mkdtemp()
-    path = join(chkpt_dir, "chkpt")
+    run_root = join(chkpt_dir, "chkpt")
     n1 = nn.Linear(1, 1)
-    c1 = [Checkpointer(networks=[n1], callbacks=[], path=path, checkpoint_interval=1)]
+    c1 = [Checkpointer(networks=[n1], callbacks=[], run_root=run_root, checkpoint_interval=1)]
```
tgolsson, line 49: Is the checkpoint used in any launchers or similar?

PR #136, opened 2023-05-30
Add more logging to ckpt/onnx/memory

jaxs-ribs:


emote/callbacks/checkpointing.py
```diff
@@ -121,6 +122,7 @@ def begin_training(self):
             net.load_state_dict(state)
         for opt, state in zip(self._opts, state_dict["optim_state_dicts"]):
             opt.load_state_dict(state)
+        logging.info(f"Loading checkpoints from {final_path}")
```
tgolsson, line 125: Print before loading the file; then finished. (So like everywhere else :D)

singhblom (approved):

emote/memory/memory.py
```diff
@@ -302,6 +302,9 @@ def add(
         observations: Dict[AgentId, DictObservation],
         responses: Dict[AgentId, DictResponse],
     ):
+        logging.info("Starting Memory export...")
+        start_time = time.time()
```
tgolsson, line 306: This should be in the if at the bottom.

tgolsson (changes requested):

tgolsson (approved):

PR #140, opened 2023-06-05
Add logging

spetravic:
Added reporting functions that can report and log additional data and histograms. Also this fixes the logging of the cumulative values, which were not being updated before.

### DIFF
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -12,8 +12,11 @@ import time
 import warnings
 
 from collections import defaultdict
+from collections.abc import Iterable
 from dataclasses import dataclass, field
 from typing import Callable, Dict, List, Mapping, Optional, Tuple, Union
+
+import torch
 
 from torch.utils.tensorboard import SummaryWriter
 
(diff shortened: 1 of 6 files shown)


emote/memory/memory.py
```diff
@@ -209,13 +212,72 @@ def add(
 
         return self._inner.add(observations, responses)
 
+    def report(
+        self,
+        metrics: Dict[str, float],
+        metrics_lists: Dict[str, List[float]],
+    ):
+        if "agent_metrics" in metrics:
+            agent_metrics = metrics.pop("agent_metrics")
```
tgolsson, line 221: Agent metrics is the one part we already handle - the agent_metrics list is set up as 
`agent_metrics = [x.metadata.info for x in observations.values()]` , which is the loop on line 196. So you can replace the logging in that loop with `self.report(obs.metadata.info, obs.metadata.info_lists)`, and do the histogram-branch below.

emote/mixins/logging.py
```diff
@@ -54,10 +55,14 @@ def log_windowed_scalar(self, key: str, value: Union[float, torch.Tensor]):
             self.windowed_scalar[key] = deque(maxlen=length)
             self.windowed_scalar_cumulative[key] = 0
 
-        if isinstance(value, torch.Tensor):
-            self.windowed_scalar[key].append(value.item())
+        if isinstance(value, Iterable):
```
tgolsson, line 58: Please update the type annotation/doc of this function too. :-)

tests/test_memory_logger.py
```diff
@@ -104,3 +104,67 @@ def test_completed(table_proxy, tmpdir):
...
+    proxy = LoggingProxyWrapper(
+        table_proxy,
+        SummaryWriter(
+            log_dir=tmpdir,
+        ),
+        2,
+    )
+
+    agent_metrics = [{"histogram:ones": [1, 1, 1, 1, 1]}, {"one": 1}]
+    metrics = {"agent_metrics": agent_metrics, "two": 2}
```
tgolsson, line 151: See note about `agent_metrics` above!

tgolsson (changes requested):
A few nits, and a note on `agent_metrics` where we can simplify!

PR #141, opened 2023-06-08
fix docs build?

tgolsson:
This fixes a few missing classes and a missing graph in the docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,7 @@ name: Deploy static content to Pages
 
 on:
   push:
-    branches: ["main", "ts/fix-docs-build"]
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -2,7 +2,7 @@ name: Deploy static content to Pages
 
 on:
   push:
-    branches: ["main"]
```
singhblom, line 5: Should this be here once I approve?

singhblom (approved):

.github/workflows/docs.yml
tgolsson: I'll revert :-)

jaxs-ribs (approved):

PR #142, opened 2023-06-09
Add Conv1d encoder and update Conv2d encoder with option to flatten

spetravic:
I added a Conv1d encoder which takes as input _(batch size, channels, length)_, or _(batch size, length, channels)_ if `channels_last=True`, following the example of the Con2d encoder. I also added the option to directly add a `nn.Flatten()` layer, instead of having to append one later on, but the _flatten_ parameter can be set to False if needed. 

Also, I updated the existing CURL callback to match the changes done to the Conv2d encoder.

### DIFF
--- a/emote/nn/layers.py
+++ b/emote/nn/layers.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple
+from __future__ import annotations
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


emote/nn/layers.py
```diff
@@ -78,6 +84,95 @@ def get_encoder_output_size(self, flatten: bool = False):
             curr_size_y = ((curr_size_y - k + 2 * p) // s) + 1
 
         out_size = (self._channels[-1], curr_size_x, curr_size_y)
-        if flatten:
-            out_size = np.prod(out_size)
+
+        if self.flatten or return_flattened:
+            flattened_out_size = np.prod(out_size)
+            return *out_size, flattened_out_size
```
singhblom, line 90: Sorry if I might have missed this, but why do we return this thing and not just `flattened_out_size`? It doesn't really make sense to me, and if we do want to return both the unflattened sizes and the flattened one, why expand the first tuple? Wouldn't it be easier to read if we just returned `(out_size, flattened_out_size)`?

spetravic: I added the unflattened one for logging purposes, but I think you are right that it doesn't make much sense to keep it. I will remove it and update the code :)

tgolsson (commented):
LGTM, but: can we get some test or example using the 1D conv encoder? :D

spetravic:
@tgolsson I added a small test to make sure that the encoder can run and the dimensions match the expected ones

tgolsson (approved):
Thanks!

PR #143, opened 2023-06-09
Agent proxy that supports multiple input keys

spetravic:
At the moment, we have agent proxies that support only one input key. These don't support the case of multiple parts in the observation, for example both features and images.

I wrote a new agent proxy that iterates through the given keys and creates a dictionary where all parts of the observation are stored with their respective keys. The policy can then handle these as needed.

Is this a good way to approach this?

### DIFF
--- a/emote/sac.py
+++ b/emote/sac.py
@@ -15,6 +15,7 @@ from emote.mixins.logging import LoggingMixin
 from emote.proxies import AgentProxy
 from emote.typing import AgentId, DictObservation, DictResponse, EpisodeState
 from emote.utils.gamma_matrix import discount, make_gamma_matrix, split_rollouts
+from emote.utils.spaces import MDPSpace
 
 
 def soft_update_from_to(source, target, tau):  # From rlkit
@@ -399,7 +400,13 @@ class VisionAgentProxy:
 class MultiKeyAgentProxy:
     """Observations are dicts that contain multiple input keys (e.g. both "features" and "images")"""
 
-    def __init__(self, policy: nn.Module, device: torch.device, input_keys: tuple):
+    def __init__(
+        self,
+        policy: nn.Module,
+        device: torch.device,
+        input_keys: tuple,
+        spaces: MDPSpace = None,
+    ):
         """Create a new proxy.
 
         Args:
@@ -411,6 +418,7 @@ class MultiKeyAgentProxy:
         self._end_states = [EpisodeState.TERMINAL, EpisodeState.INTERRUPTED]
         self.device = device
         self.input_keys = input_keys
+        self._spaces = spaces
 
     def __call__(
         self, observations: dict[AgentId, DictObservation]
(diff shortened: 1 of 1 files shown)


tgolsson:
Very clean, but I'm wondering how a dictionary input for a policy works e.g. with ONNX exporters etc. I found it much easier to let the `input_keys` just represent argument ordering. So your policy can be `def forward(self, a, b, c)` and your input keys tell the proxy what data goes with a, b, c.

singhblom (approved):
LGTM!

singhblom:
Wait Tom, just to make sure I understand what you're saying. You would like to see a 
```
actions = self.policy(dict_tensor_obs)[0].detach().cpu().numpy()
```
changed to 
```
actions = self.policy(**dict_tensor_obs)[0].detach().cpu().numpy()
```

but apart from that mostly things work fine?

spetravic:
> Very clean, but I'm wondering how a dictionary input for a policy works e.g. with ONNX exporters etc. It would also require the policy to handle the names being able to be different. I found it much easier to let the `input_keys` just represent argument ordering. So your policy can be `def forward(self, a, b, c)` and your input keys tell the proxy what data goes with a, b, c.

That makes sense. It would also help with readability in the policy, as the function signature will explicitly show what it expects to receive.

tgolsson:
@singhblom No, not even that. That requires the policy function to be defined using the keys of the agent proxy. That in turn makes the Python code determine the name of the inputs sent from the game. What I want to see is:

```
tensor_obs = []
for key in self._input_keys: 
     inputs.append(...)

actions = self.policy(*tensor_obs)[0].detach().cpu().numpy()
```

Then, the function can be `def forward(self, a, b, c)`, while the game sends `{'foobar': [...], 'image': [ ... ], 'hidden': [ ... ] }`. With input_keys being set to `('image', 'foobar', 'hidden',)` we'd map `a=image, b=foobar, c=hidden`.

tgolsson:
Yepp! And with the current setup; if I send `feature_data` instead of `features` it all goes up in flames everywhere. :-) No reason for it, and using positionals at the internal layer solves that - like the FeatureAgentProxy. You could put the same code I just posted in the feature agent proxy, actually, and it'd still work because we have 1 positional - 2 with epsilon - already. No need for dictionaries, unless we make sure *no part* has a hardcoded name when reading from it. 

Another way of thinking it is that I'm *remapping* the names. Again, `{'image': 'a', 'foobar': 'b', 'hidden': 'c'}`. That's also fine. I could say `'feature_data': 'a'` if that's what my data generator is generating. `**tensor_dict_obs` is a bug waiting to happen - `input_names` was added because some code assumed the features were called `obs`. That requires a separate policy and q network where that name *is in the function signature*. With `**`, you cannot use that network for any other input name at that point (without remapping explicitly!).

emote/sac.py
```diff
@@ -433,6 +441,12 @@ def __call__(
                     for agent_id in active_agents
                 ]
             )
+
+            if self._spaces is not None:
+                shape = (np_obs.shape[0],) + self._spaces.state.spaces[input_key].shape
+                if shape != np_obs.shape:
+                    np_obs = np.reshape(np_obs, shape)
```
singhblom, line 448: I see. LGTM.

singhblom (approved):

tgolsson (approved):
LGTM with all the discussion had elsewhere :D

PR #144, opened 2023-06-13
Simona/fix logging

spetravic:
1. Fixes my own code added in #140 for histogram logging, as it was missing `.items()` to properly iterate through the dictionary when sending to the writer. Also, PyTorch complains about getting a deque as it expects a numpy array, torch tensor, or caffe2 blob name, so now the deque is converted to a numpy array.
2. Fixes wrong key splitting in _log_windowed_scalar_ for overriding the window size.

### DIFF
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -12,11 +12,12 @@ import os
 import time
 import warnings
 
-from collections import defaultdict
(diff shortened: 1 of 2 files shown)


tgolsson (approved):
LGTM!

PR #146, opened 2023-06-21
updating the checkpointer and checkpoint loader

AliGhadirzadeh:
Made some updates to the Checkpointer and Checkpointloader as follows:

* Removed the extra input arguments for loading networks and optimizers. Now, you can only load a list of callbacks. However, added an additional input to the Checkpointloader class. When set to True, this input instructs the loader to exclusively load the neural network parameters, disregarding the state_dict of other callbacks. This feature is useful in scenarios where you need to copy a network from one callback to another or when you prefer not to load other callback parameters.

Both Checkpointer and Checkpointloader rely on a list of callbacks rather than a dictionary. To ensure the correct loading of callbacks, we store a corresponding list of names when saving the checkpoints. These names serve as references during the loading process. The names are loaded from the callback names. Callbacks without a name field are discarded. An error will be raised in case repetitive names are given.

### DIFF
--- a/emote/callback.py
+++ b/emote/callback.py
@@ -211,7 +211,7 @@ class Callback(metaclass=CallbackMeta):
     def state_dict(self) -> Dict[str, Any]:
         return {}
 
-    def load_state_dict(self, state_dict: Dict[str, Any]):
+    def load_state_dict(self, state_dict: Dict[str, Any], *args, **kwargs):
         pass
 
 
--- a/emote/callback.py
+++ b/emote/callback.py
@@ -211,7 +211,13 @@ class Callback(metaclass=CallbackMeta):
     def state_dict(self) -> Dict[str, Any]:
         return {}
 
-    def load_state_dict(self, state_dict: Dict[str, Any], *args, **kwargs):
+    def load_state_dict(
+        self,
+        state_dict: Dict[str, Any],
+        load_network: bool = True,
+        load_optimizer: bool = True,
+        load_hparams: bool = True,
+    ):
         pass
 
 
--- a/emote/callbacks/loss.py
+++ b/emote/callbacks/loss.py
@@ -52,11 +52,16 @@ class LossCallback(LoggingMixin, Callback):
             state["network_state_dict"] = self.network.state_dict()
         return state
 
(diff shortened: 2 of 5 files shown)


emote/callback.py
```diff
@@ -211,7 +211,7 @@ def end_training(self, *args, **kwargs):
     def state_dict(self) -> Dict[str, Any]:
         return {}
 
-    def load_state_dict(self, state_dict: Dict[str, Any]):
+    def load_state_dict(self, state_dict: Dict[str, Any], *args, **kwargs):
```
tgolsson, line 214: I don't think this has to be generic! We have previously used three different flags and I'd prefer to maintain that flow:

* load_weights
* load_optimizers
* load_hparams - this is a bit badly named, but essentially inf-steps and other things like that that aren't in either of the above categories! 

I think that would make more sense than having flags that are "only x". What if I want only optimizers? Only schedules? Weights and optimizers but no schedules? :-)

tgolsson (changes requested):
A minor change request to ensure we don't build ourselves into a corner! I think exclusive flags get quite messy over time, so think clear opt ins/outs on individual parts make more sense. I think that will make some other code cleaner too. :-)

singhblom (approved):
This looks clean and good to me.

emote/callbacks/loss.py
```diff
@@ -52,12 +52,21 @@ def state_dict(self):
...
+    ):
+        if self.network and load_weights:
             self.network.load_state_dict(state_dict.pop("network_state_dict"))
-        super().load_state_dict(state_dict)
+
+        if self.optimizer and load_optimizers:
+            self.optimizer.load_state_dict(state_dict.pop("optimizer_state_dict"))
+
+        if load_hparams:
+            super().load_state_dict(state_dict)
```
tgolsson, line 69:
```suggestion
        super().load_state_dict(state_dict, load_weights, load_optimizers, load_hparams)
```

We should always call this function, and always pass the flags. :-)

AliGhadirzadeh: Thanks for it! it is fixed now.

tgolsson (approved):
Awesome! LGTM.

PR #147, opened 2023-06-27
Fix alpha loss load_state_dict fn signature

klashenriksson:


AliGhadirzadeh (approved):

emote/sac.py
```diff
@@ -296,10 +296,23 @@ def state_dict(self):
...
+    def load_state_dict(
+        self,
+        state_dict: Dict[str, Any],
+        load_weights: bool = True,
+        load_optimizer: bool = True,
+        load_hparams: bool = True,
+    ):
+        saved_ln_alpha = state_dict.pop("network_state_dict")
+
+        self.ln_alpha.requires_grad_(False)
```
singhblom, line 308: NICEEEEE! I was stupid! This is tricky. Could we make sure these things always work correctly in some way? Good find.

singhblom (approved):

emote/sac.py
klashenriksson: Not sure, would be nice to have a find of catching these, it is indeed tricky and very easy to do

tgolsson: UNIT TESTS. UNIT TESTS. UNIT TESTS. 🪧

singhblom: Unit tests are nice but a bit orthogonal to what I would like to see. This seems like it could be a class of problem that it is easy for implementers to forget. A unit test here solves this instance. What I'd like is to maybe find all tensors on a callback before `load_state_dict`, load the dict, find all the tensors after, and verify that they have the same identity. So try to enforce the invariant instead of testing each case. Could we maybe find all members of type tensor before `load_state_dict` and then verify their identities after in `CheckpointLoader`'s `begin_training`?

PR #149, opened 2023-06-29
Update windowed scalar (again)

spetravic:
After fixing the overwriting problem in #148, I found that by logging using the split key we cause problems in other places where we request data from the dictionary. By the same logic as #148, when we want to get logged data we would request the original key, e.g. "windowed[100]:score" which doesn't exist in the dictionary, so we don't get any value back. Anything that depends on these values is then broken. 

Instead of splitting the key in a bunch of other places, this code uses the original key in the dictionary and then splits it only for logging to the writer.

### DIFF
--- a/emote/mixins/logging.py
+++ b/emote/mixins/logging.py
@@ -50,7 +50,7 @@ class LoggingMixin:
         if key not in self.windowed_scalar:
             # we allow windowed[100]:some_key/foobar to override window size
             if "windowed[" in key:
-                p, k = key.split(":")
+                p = key.split(":")[0]
                 length = int(p.split("[")[1][:-1])
             else:
                 length = self._default_window_length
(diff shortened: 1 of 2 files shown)


emote/mixins/logging.py
```diff
@@ -47,15 +47,14 @@ def log_windowed_scalar(
...
-            p, k = key.split(":")
-            length = int(p.split("[")[1][:-1])
-            key = k
-        else:
-            length = self._default_window_length
-
         if key not in self.windowed_scalar:
+            # we allow windowed[100]:some_key/foobar to override window size
+            if "windowed[" in key:
+                p, k = key.split(":")
```
tgolsson, line 53: k isn't actually used anymore, is it?

spetravic: You are right, I forgot to update that line. Changed it to `p = key.split(":")[0]` :)

tgolsson (approved):
LGTM!

PR #151, opened 2023-07-05
Remove _inf_step from tb logs

singhblom:
Why do we suffix this with inf-step? It gets very noisy and makes it much harder to compare the graphs between cog and emote. Could we remove it while we're debugging at least?

### DIFF
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -273,7 +273,7 @@ class LoggingProxyWrapper(TableMemoryProxyWrapper, LoggingMixin):
                 self.completed_episodes,
             )
 
-        suffix = "inf_step"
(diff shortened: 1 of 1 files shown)


tgolsson:
I'm fine with removing it completely, would make the code much nicer! I just followed the pattern from other loggers.

singhblom:
Yeah, we have the bp_step thing as well. I would like to keep that suffix for now, since cog logs both bp_samples and bp_steps and if we keep it the logs en up in the same plot for now. But the inf could be skipped for now.

After we've gotten emote to parity we can talk about if we should have any suffixes at all. I think they can all go. @spetravic, opinions?

spetravic (approved):

PR #152, opened 2023-07-05
Orthogonal init for conv as well

singhblom:


singhblom:
I think our convolutions are bias-free. Lemme double check though.

klashenriksson:
hmm now that i look at this, should we maybe do that check for linear layers as well? just to not error out if we ever use non-biased linear layers

klashenriksson (approved):

PR #153, opened 2023-07-05
add PDM 2.6 and 2.7 to CI image

tgolsson:
Want to look at upgrading our plugins quite soon, this is a good initial step so I can start testing on CI. Also adds `--pip-args='--no-cache-dir'` which reduces image size a tiny bit.

### DIFF
--- a/.buildkite/Dockerfile
+++ b/.buildkite/Dockerfile
@@ -20,7 +20,6 @@ RUN mkdir -p /gcloud/ && \
 RUN pip install \
 	--no-cache-dir \
 	pipx && \
-	pipx install --pip-args='--no-cache-dir' pdm~=2.4.0 && \
 	pipx install --pip-args='--no-cache-dir' pdm~=2.3.0 --suffix 23 && \
 	pipx install --pip-args='--no-cache-dir' pdm~=2.4.0 --suffix 24 && \
 	pipx install --pip-args='--no-cache-dir' pdm~=2.5.0 --suffix 25 && \
@@ -30,7 +29,6 @@ RUN pip install \
 	pdm24 --version && \
 	pdm25 --version && \
 	pdm26 --version && \
-	pdm27 --version && \
-	pdm --version
+	pdm27 --version
 
 ENTRYPOINT bash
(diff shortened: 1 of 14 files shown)


singhblom (approved):

.buildkite/Dockerfile
```diff
@@ -20,13 +20,17 @@ RUN mkdir -p /gcloud/ && \
 RUN pip install \
 	--no-cache-dir \
 	pipx && \
-	pipx install pdm~=2.4.0 && \
-	pipx install pdm~=2.3.0 --suffix 23 && \
-	pipx install pdm~=2.4.0 --suffix 24 && \
-	pipx install pdm~=2.5.0 --suffix 25 && \
+	pipx install --pip-args='--no-cache-dir' pdm~=2.4.0 && \
```
ivarg, line 23: Perhaps there's value in not having a default version and instead force users to be explicit about which version they want to use

ivarg (approved):

.buildkite/publish-docs.sh
```diff
@@ -3,18 +3,16 @@ set -eo pipefail
...
 source .buildkite/install-repo.sh
 
-echo --- Initializing gcloud
-
-
 echo --- Building docs
 pushd docs
 EXIT_CODE=0
-pdm run make deploy || EXIT_CODE=$?
+PDM=${PDM_COMMAND:1:-1} ${PDM_COMMAND:1:-1} run make deploy  || EXIT_CODE=$?
```
ivarg, line 9: An ignorant question: what does ${PDM_COMMAND:1:-1} do? I thought it produced a substring with the first and last characters removed

ivarg (approved):

.buildkite/publish-docs.sh
tgolsson: It does! Buildkite has the weird side-effect of hard-quoting all env variables set in the pipeline.yml (and I don't want to search replace in a million files for pdm25 whenever we upgrade). So if we try to run `$PDM_COMMAND` it actually runs `"pdm25" some command`, but `"pdm25"` isn't a program - only `pdm25` is.

PR #156, opened 2023-07-05
Change check for bias initialization

spetravic:
The current check raises RuntimeError: Boolean value of Tensor with more than one value is ambiguous


klashenriksson (approved):
Nice

singhblom (approved):
I stupid.

PR #158, opened 2023-07-07
add optional per-parameter logging for loss callbacks

klashenriksson:
This PR adds the ability to log per-parameter information, similar to what we had in Cog. It is opt-in and off by default.

See attached screenshot for examples of how it can look like.
per_param

### DIFF
--- a/emote/callbacks/loss.py
+++ b/emote/callbacks/loss.py
@@ -1,7 +1,8 @@
 from typing import Any, Dict, Optional
 
-from torch import Tensor, nn, optim
 import torch
+
+from torch import Tensor, nn, optim
 from torch.optim import lr_scheduler
 
 from emote.callback import Callback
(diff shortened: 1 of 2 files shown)


emote/callbacks/loss.py
```diff
@@ -44,6 +61,35 @@ def backward(self, *args, **kwargs):
         self.log_scalar(f"loss/{self.name}_loss", loss)
         self.log_scalar(f"loss/{self.name}_gradient_norm", grad_norm)
 
+        if self._log_per_param_weights or self._log_per_param_grads:
+            self.log_per_param_weights_and_grads()
+
+    def log_per_param_weights_and_grads(self):
+        def _friendly_shape_str(shape):
+            return str(list(shape)).replace("[", "").replace("]", "").replace(", ", "_")
```
tgolsson, line 69: I'd define this outside the function, avoids recreating the function each time. :-)

klashenriksson: Ah I did not know that was how it worked. Will fix

singhblom (approved):
Nice!

PR #159, opened 2023-08-18
Ali/genrl pr

AliGhadirzadeh:
This is the implementation of GenRL algorithm in Emote.


emote/nn/layers.py
```diff
@@ -174,3 +175,159 @@ def get_encoder_output_size(self):
...
+        super().__init__()
+        self.device = device
+
+        self.input_size = input_size
+        self.output_size = output_size
+        self._freeze_bn = freeze_bn
+
+        num_layers = len(hidden_sizes)
+        batch_norm = (
+            FrozenBatchNorm1d(hidden_sizes[0])
```
singhblom, line 199: Does the frozen BN need any extra code to restore correctly from checkpoints?

emote/sac.py
```diff
@@ -126,6 +126,7 @@ def __init__(
         )
 
     def begin_batch(self, next_observation, rewards, masks):
+        # print(next_observation)
```
singhblom, line 129: This looks like a forgotten debug statement. Remove?

singhblom (approved):
Really nice overall! Let's get it merged so more people can play with it.

emote/algos/genrl/nn.py
```diff
@@ -0,0 +1,57 @@
...
+        self.action_encoder = action_encoder
+        self.num_action = num_actions
+
+    def forward(self, action, obs):
+
+        if len(action.shape) == 1:
+            action = action.unsqueeze(dim=0)
+
+        if len(obs.shape) == 1:
+            obs = obs.unsqueeze(dim=0)
```
tgolsson, line 49: 😕 Again, cursed

emote/algos/genrl/nn.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if action.shape[1] != self.num_action:
+            latent_action = self.action_encoder(
+                action,
+                obs,
+            )
+            return super().forward(latent_action, obs)
```
tgolsson, line 56: Could do `action = self.action_encoder(...)` and keep the branched block smaller.

emote/algos/genrl/nn.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def forward(self, action, obs):
+
+        if len(action.shape) == 1:
+            action = action.unsqueeze(dim=0)
+
+        if len(obs.shape) == 1:
+            obs = obs.unsqueeze(dim=0)
+
+        if action.shape[1] != self.num_action:
```
tgolsson, line 51: This seems... cursed as well. Why isn't the encoder a wrapper around this module instead?

emote/algos/genrl/proxies.py
```diff
@@ -0,0 +1,46 @@
...
+        self._input_key = input_key
+
+    def add(
+        self,
+        observations: Dict[AgentId, DictObservation],
+        responses: Dict[AgentId, DictResponse],
+    ):
+        updated_responses = {}
+        for agent_id, response in responses.items():
+            actions = np.array(response.list_data["actions"])
```
tgolsson, line 32: hardcoded string here; should probably make it like input_key

emote/algos/genrl/proxies.py
```diff
@@ -0,0 +1,46 @@
...
+            if np.size(actions) == 0:
```
tgolsson, line 33: Hmm, so why would we have an empty actions?

emote/algos/genrl/proxies.py
```diff
@@ -0,0 +1,46 @@
...
+        self,
+        table: Table,
+        encoder: nn.Module,
+        minimum_length_threshold: Optional[int] = None,
+        use_terminal: bool = False,
+        input_key: str = "obs",
+    ):
+        super().__init__(table, minimum_length_threshold, use_terminal)
+        self.encoder = encoder
+        self._input_key = input_key
```
tgolsson, line 23: Curve-ball: If we wanted to do a hidden-state style memory network (GRU, LSTM, ...) could this support it, or what would need to change?

emote/algos/vae/nn.py
```diff
@@ -0,0 +1,129 @@
...
+            observation = observation.to(self.device)
+            condition = self.condition_fn(observation)
+
+        sample = self.decoder.forward(latent, condition)
+
+        return sample
+
+    def load_state_dict(self, state_dict, strict=True):
+        model_dict = self.state_dict()
+        new_state_dict = {k: v for k, v in state_dict.items() if k in model_dict}
```
tgolsson, line 67: Nice "hack" :D

emote/algos/genrl/proxies.py
```diff
@@ -0,0 +1,46 @@
...
+        for agent_id, response in responses.items():
+            actions = np.array(response.list_data["actions"])
+            if np.size(actions) == 0:
+                updated_responses.update({agent_id: response})
+            else:
+                actions = torch.from_numpy(actions).unsqueeze(dim=0).to(torch.float)
+                obs = torch.from_numpy(
+                    observations[agent_id].array_data[self._input_key]
+                )
+                obs = obs.unsqueeze(dim=0).to(torch.float)
```
tgolsson, line 40: All these squeeze/unsqueeze feel a bit ugly. I wonder if we can make it a bit nicer somehow.

emote/algos/vae/vae.py
```diff
@@ -0,0 +1,82 @@
...
+        self.conditioning_func = conditioning_func
+        self._input_key = input_key
+
+    def loss(self, observation, actions):
+
+        condition = self.conditioning_func(observation[self._input_key])
+        samples, dist_mean, dist_log_std, _ = self.vae.forward(actions, condition)
+        loss, info = self.vae.loss(actions, samples, dist_mean, dist_log_std)
+        self.log_scalar("genrl/restore_loss", torch.mean(info["restore_loss"]))
+        self.log_scalar("genrl/kl_loss", torch.mean(info["kl_loss"]))
```
tgolsson, line 80: genrl -> vae?

emote/algos/vae/vae.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(
+        self,
+        *,
+        vae: VariationalAutoencoder,
+        opt: optim.Optimizer,
+        lr_schedule=None,
+        max_grad_norm: float = 10.0,
+        name: str = "vae",
+        data_group: str = "default",
+        input_key: str = "obs",
```
tgolsson, line 59: input_key => input_keys :-) (and list, obviously!). To support images etc later.

emote/nn/misc.py
```diff
@@ -0,0 +1,44 @@
+import torch.nn.functional as F
+
+from torch import Tensor, nn
+
+
+class FrozenBatchNorm1d(nn.BatchNorm1d):
```
tgolsson, line 6: The issue and reason for this was that it was randomly flipping training/not training, right?

emote/nn/layers.py
```diff
@@ -174,3 +175,159 @@ def get_encoder_output_size(self):
             out_size = np.prod(out_size)
 
         return out_size
+
+
+class FullyConnectedEncoder(nn.Module):
```
tgolsson, line 180: I think we've argued back and forth whether we should have encoders etc, or if Emote should prefer lower-level building blocks only.

tgolsson: Maybe worth making a call here and then sticking to it. cc @singhblom

emote/nn/layers.py
```diff
@@ -174,3 +175,159 @@ def get_encoder_output_size(self):
...
+                condition (torch.Tensor): batch x condition_size
+
+            Returns:
+                tuple(torch.Tensor, torch.Tensor): the mean (batch x data_size)
+                    and log_std (batch x data_size)
+        """
+        assert len(data.shape) == 2
+        if condition is not None:
+            assert len(condition.shape) == 2
+            data = torch.cat((data, condition), dim=1)
```
tgolsson, line 248: This seems like a such a specific thing... Could we not just compose this with a wrapper instead?

tgolsson (changes requested):
Looks mostly good! Let's clean these up and fix algos -> algorithms at least.

emote/nn/layers.py
singhblom: oof. Yes, we have. I think my opinion here is actually that things that are this big should probably be in the protocol in erupt.

emote/nn/layers.py
AliGhadirzadeh: I believe It will inherit it from the _NormBase class. The only difference to the base class is that it ensures no further updates in the forward pass.

emote/sac.py
AliGhadirzadeh: fixed

emote/nn/layers.py
AliGhadirzadeh: Sounds very right to me :)

AliGhadirzadeh: Fixed!

emote/algos/genrl/proxies.py
AliGhadirzadeh: We construct a new MDP with a different action space, which we call it latent action space. The proxy first maps actions into the latent action space and then stores the encoded actions instead of the original ones. As long as there is a 1-1 mapping between the original actions and the latent actions, there is no added complexity using genrl. This is the case (1-1 mapping) when using only action data for the mapping. It is also fine when conditioning on observation since we only get the proprioceptive data (we do not deal with partial observations).

emote/nn/misc.py
AliGhadirzadeh: Yes, this is correct

AliGhadirzadeh: The frozen_bn is only needed for the encoder and decoder models which are moved to erupt. Hence, I move this one also to erupt

tgolsson: :+1: We can probably remove it completely, @spetravic and I looked at a similar issue caused by the ONNX exporter, and it now copies the network before setting the export state. We should maybe also make copies that are only used by the data generation loop that's always in infer mode.

emote/algos/genrl/nn.py
```diff
@@ -0,0 +1,57 @@
...
+        self.num_actions = num_actions
+        self.decoder = decoder
+
+    def forward(self, obs: Tensor, epsilon: Tensor = None):
+        obs = obs.to(torch.float)
+        if len(obs.shape) == 1:
+            obs = obs.unsqueeze(dim=0)
+
+        if epsilon is not None:
+            epsilon = epsilon[:, : self.num_actions]
```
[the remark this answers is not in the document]
AliGhadirzadeh: @tgolsson I've removed all squeeze and unsqueeze things. The only place that we need to unsqueeze is in FrozenBatchNorm. And, if I correctly understood you, the problem with the regular BatchNorm is fixed now, and we can remove FrozenBatchNorm completely soon (if not now). 
However, I cannot fix the if-statement for epsilon. I must reduce the size of the tensor (epsilon) when it is not None. Otherwise, I'll get error later in the code when epsilon is multiplied by the policy std output. Please note that epsilon is generated for the original action space and with GenRL the action space is reduced to the size of the latent space. One possible solution can be to configure the code such that it generates epsilon with the right size. But, I do not know how. Any idea?

emote/algos/genrl/nn.py
AliGhadirzadeh: fixed

emote/algos/genrl/nn.py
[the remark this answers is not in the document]
tgolsson: Makes sense! A comment around that might be enough, but I'm not sure I understand the problem exactly. Is the distribution not per-action, but rather per-latent and the output action is derived directly from that?

emote/algos/genrl/nn.py
AliGhadirzadeh: Fixed!

emote/algos/genrl/nn.py
AliGhadirzadeh: The two classes, GenRLPolicy and GenRLQNet, are removed from the newer implementation.

emote/algos/genrl/proxies.py
AliGhadirzadeh: They are all fixed

emote/algos/vae/vae.py
AliGhadirzadeh: Thanks for it! fixed

emote/algos/vae/vae.py
AliGhadirzadeh: It is only being used for conditioning and conditioning should be only done based on propioceptives and not images. Currently, proprioceptives are a part of features.

emote/algos/genrl/proxies.py
AliGhadirzadeh: fixed

AliGhadirzadeh:
@tgolsson and @singhblom 
I updated the code by changing the structure a bit. There is no need for GenRLPolicy and GenRLQNet anymore with the new implementation. We are re-using most of SAC code by only changing the action space into the latent space. A PolicyWrapper is implemented which maps the latent actions generated by the policy into the actions in the original action space and is passed to the AgentProxy. Please take a look at the code and let me know if you have any other comments.

singhblom (approved):

AliGhadirzadeh:
Poke @tgolsson

emote/algorithms/genrl/wrappers.py
```diff
@@ -0,0 +1,134 @@
...
+        self.latent_size = decoder.latent_size
+        self.decoder = decoder
+        self.policy = policy
+
+    def forward(self, obs: Tensor, epsilon: Tensor = None):
+        # we need to discard the extra dimensions of epsilon.
+        # the input epsilon is given for the original action space
+        # however, the policy outputs latent actions.
+        if epsilon is not None:
+            epsilon = epsilon[:, : self.latent_size]
```
tgolsson, line 127: I'll make a note to augment the ONNX exporter so it can handle this; this feels like it could end up in other places too.

tgolsson (approved):
LGTM! Great work, and much cleaned up :-)

PR #160, opened 2023-08-22
setup multi-python image

tgolsson:
Updates the base CI runner image to use pyenv to provide multiple python version using pyenv. 

This should allow us to reuse this image for all our repositories. See `install-repo.sh` for how to initialize pyenv and select a Python version.

### DIFF
--- a/.buildkite/Dockerfile
+++ b/.buildkite/Dockerfile
@@ -23,6 +23,12 @@ RUN curl https://pyenv.run | bash && \
 	python --version && \
 	pyenv exec pip --version
 
+RUN mkdir -p /gcloud/ && \
+	curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-408.0.1-linux-x86_64.tar.gz && \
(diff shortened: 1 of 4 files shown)


.buildkite/Dockerfile
```diff
@@ -1,34 +1,50 @@
...
 	curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-408.0.1-linux-x86_64.tar.gz && \
 	tar -xf google-cloud-cli-408.0.1-linux-x86_64.tar.gz -C /gcloud/ && \
 	/gcloud/google-cloud-sdk/install.sh --quiet --usage-reporting false && \
 	rm google-cloud-cli-408.0.1-linux-x86_64.tar.gz
 
-RUN pip install \
-	--no-cache-dir \
-	pipx && \
+RUN eval "$(pyenv init --path)" && \
+	eval "$(pyenv init -)" && \
```
ivarg, line 33: Is `pyenv init -` meant to be run twice?

ivarg (approved):
This is great!

.buildkite/Dockerfile
tgolsson: It has to be run with `-` and `--path` apparently.

PR #161, opened 2023-08-24
Simona/onnx

spetravic:
Creates a copy of the policy for exporting to avoid the interference of the training and inference loops


tgolsson (approved):
LGTM! @AliGhadirzadeh I wonder if a similar hack would work with the forward/backward issue you had with BatchNorm?

PR #162, opened 2023-08-24
Simona/add conditions

spetravic:
- In the Q target updates, terminal masking was used by default, and now the option is added to deactivate that when needed (+ some updates to the docstring)
- In the CURL loss calculation, the max was subtracted by default, as in the original paper. When used together with the temperature variant (which is different from the original paper), it led to problems with the gradients and loss not changing. A condition is added now to only remove the max if not using temperature.

### DIFF
--- a/emote/nn/curl.py
+++ b/emote/nn/curl.py
@@ -162,7 +162,7 @@ class CurlLoss(LossCallback):
         use_per_image_mask_size: bool = False,
         use_fast_augment: bool = False,
         use_projection_layer: bool = True,
-        augment_anchor_and_pos: bool = False,  # disabling this saves some computation and doesn't seem to have any adverse effects.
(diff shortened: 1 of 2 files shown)


emote/nn/curl.py
```diff
@@ -162,7 +162,7 @@ def __init__(
         use_per_image_mask_size: bool = False,
         use_fast_augment: bool = False,
         use_projection_layer: bool = True,
-        augment_anchor_and_pos: bool = True,  # disabling this saves some computation and doesn't seem to have any adverse effects.
+        augment_anchor_and_pos: bool = False,  # disabling this saves some computation and doesn't seem to have any adverse effects.
```
tgolsson, line 165: Is this default necessary to change? The other ones are bugfixes or opt-in, this might change semantics where both options are valid.

spetravic: Not necessary for sure!

tgolsson (approved):
LGTM!

PR #167, opened 2023-09-05
Upgrade memory serialization workflows

tgolsson:
This upgrades our memory serialization workflows to avoid pickling.

This *also* removes the identity shenanigans we do when loading memories right now. Clients have to ensure loaded memories don't overlap with fresh memories.


emote/memory/table.py
```diff
@@ -36,12 +61,16 @@ def add_sequence(self, identity: int, sequence):
         """add a fully terminated sequence to the memory"""
         ...
 
-    def store(self, path: str) -> bool:
+    def store(
+        self, path: str, version: TableSerializationVersion | None = None
```
klashenriksson, line 65: why do we need to allow this being None? Cant we default to LATEST?

klashenriksson: (same comment for restore)

klashenriksson (changes requested):

emote/memory/table.py
tgolsson: For restore there's a semantic difference (override vs "use the file"). Here, yes. Good catch 👍

klashenriksson: Right of course, sorry that makes sense. Nice.

emote/memory/column.py
```diff
@@ -18,6 +20,31 @@ class Column:
...
+            dtype_name = self.dtype.name
+        else:
+            dtype_name = self.dtype.__name__
+
+        return {
+            "shape": self.shape,
+            "dtype": dtype_name,
+        }
+
+    def configure(self, config):
```
singhblom, line 35: Why not name these two `state` and `load_state`?

emote/memory/fifo_strategy.py
```diff
@@ -43,6 +48,20 @@ def post_import(self):
...
+            "identities": list(self._identities),
+            "sequence_lengths": list(self._sequence_lengths.items()),
+        }
+
+    def load_state(self, state: dict):
+        """Load the strategy from a dictionary"""
+        self._identities = deque(state["identities"])
+        self._sequence_lengths = dict(state["sequence_lengths"])
+
+        self._in_simple_import = True
```
singhblom, line 63: This is confusing to me. Where is this unset? I can't figure out this logic. It feels like this flag should be handled by `begin_simple_import` and `end_simple_import`. Why does it need to be handled explicitly here? And why does that happen at the end of load_state?

emote/memory/table.py
```diff
@@ -228,11 +266,178 @@ def _eject_count(self, count: int):
...
+                    end = start + size
+                    reassembled[key] = loaded_data[key][start:end]
+
+                self._add_sequence_internal(identity, reassembled)
+
+            self._sampler.end_simple_import()
+            self._ejector.end_simple_import()
+
+    def _store_legacy(self, path: str) -> bool:
         """Persist the whole table and all metadata into the designated name"""
```
singhblom, line 440: Could we deprecation warn here maybe?

singhblom (changes requested):
Nice work, this seems much easier and more consistent to work with. I left a few comments, mostly nitpicks and naming things.

emote/memory/column.py
tgolsson: Because it's not state, just configuration. :-)

emote/memory/fifo_strategy.py
tgolsson: It indeed should not! Originally `in_simple_import` was managed by each strategy, not through the base class.

emote/memory/table.py
```diff
@@ -14,6 +21,24 @@
 from .strategy import EjectionStrategy, SampleStrategy
 
 
+logger = logging.getLogger(__name__)
+
+
+class TableSerializationVersion(enum.Enum):
+    """The version of the memory serialization format."""
+
+    Legacy = 0
```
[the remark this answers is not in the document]
tgolsson: Legacy wasn't versioned, but to avoid making "unversioned" = -1 I went for `unversioned = 0`. Does that make sense? So V1 is the first actually versioned protocol.

emote/memory/table.py
tgolsson: Since it's now an internal function, I'm not sure if that makes sense? All existing code will be automatically upgraded to use the new format, and if you need the old format you have to explicitly opt-in.

emote/memory/column.py
singhblom: I think it makes me a bit sad to follow exactly the same pattern we do for every single pytorch layer and emote callback but use different names for things.

tgolsson: Fair point, fixed.

emote/memory/table.py
tgolsson: Added deprecation warnings.

tgolsson: image

AliGhadirzadeh (approved):
Thanks for the upgrade. It looks great to me.

singhblom (approved):

klashenriksson (approved):
Sweet!

PR #168, opened 2023-09-08
add `OnnxExporter.add_metadata`

tgolsson:
This lets downstream users add things like configuration and hyperparameters.

We should probably ensure we can override the basic ONNX metadata as well at some point, but not sure about the API so ignoring for now.
```
                model_proto.producer_name = "emote"
                model_proto.domain = "dev.embark.ml"
                model_proto.producer_version = self.version_tag
                model_proto.model_version = model_version
                model_proto.doc_string = "exported via Emote checkpointer"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


singhblom (approved):

klashenriksson (approved):
Nice!

PR #169, opened 2023-09-13
Add "latest_checkpoint" path to state_dict

jaxs-ribs:
Doesn't quite mirror [this cog PR](https://github.com/EmbarkStudios/cog/pull/731/files), but ensures we have the path, as discussed with @tgolsson. Please let me know in case I misunderstood something.

### DIFF
--- a/emote/callbacks/checkpointing.py
+++ b/emote/callbacks/checkpointing.py
@@ -74,8 +74,8 @@ class Checkpointer(Callback):
         self._checkpoint_index += 1
 
         subset = {
-            key: value for key, value in state_dict["training_state"].items()
-            if key not in ["bp_step", "bp_samples"]
+            "latest_checkpoint": state_dict["training_state"]["latest_checkpoint"],
+            "checkpoint_index": state_dict["training_state"]["checkpoint_index"]
         }
         return subset
 
--- a/emote/callbacks/checkpointing.py
+++ b/emote/callbacks/checkpointing.py
@@ -75,7 +75,7 @@ class Checkpointer(Callback):
 
         return {
             "latest_checkpoint": state_dict["training_state"]["latest_checkpoint"],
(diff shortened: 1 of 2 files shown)


singhblom (approved):

tgolsson:
This is OK to add; but now we're writing into the checkpoint what path the checkpoint went to. As noted, we want to give this information out to other parts of the codebase; so we need to return that information in a dict. Then it gets merged with the rest of the states in the big state dict in the trainer:

```py3
return {
    "last_checkpoint_path": final_path,
    "last_checkpoint_index": self._checkpoint_index - 1,
}
```

There should be unit tests for the checkpointer, so you can write a unit test for that being in https://github.com/EmbarkStudios/emote/blob/main/emote/trainer.py#L32 after a checkpoint has been saved.

jaxs-ribs:
> This is OK to add; but now we're writing into the checkpoint what path the checkpoint went to. As noted, we want to give this information out to other parts of the codebase; so we need to return that information in a dict. Then it gets merged with the rest of the states in the big state dict in the trainer:
> 
> ```python
> return {
>     "last_checkpoint_path": final_path,
>     "last_checkpoint_index": self._checkpoint_index - 1,
> }
> ```
> 
> There should be unit tests for the checkpointer, so you can write a unit test for that being in https://github.com/EmbarkStudios/emote/blob/main/emote/trainer.py#L32 after a checkpoint has been saved.

I return the state now and added a test that makes sure the key is in. Should the checkpoint index really be with a -1 though?

tgolsson:
> I return the state now and added a test that makes sure the key is in. Should the checkpoint index really be with a -1 though?

We shouldn't return all callback states; though. That has no use for anyone else (or shouldn't have...). Only the data we want to share with other callbacks.  The -1 was because it was returning after the increment.

tgolsson (approved):

PR #175, opened 2023-10-09
Implement DQN and CartPole Test

jaxs-ribs:
Adds the DQN Algorithm, as well as a corresponding CartPole test. The reward usually goes over 200 after around 25K episodes. 

Also moving sac.py and dqn.py to algorithms, and refactoring accordingly.


emote/algorithms/dqn.py
```diff
@@ -0,0 +1,199 @@
...
+from emote.callbacks.loss import LossCallback
+from emote.mixins.logging import LoggingMixin
+from emote.proxies import AgentProxy
+from emote.algorithms.sac import soft_update_from_to
+from emote.typing import AgentId, DictObservation, DictResponse, EpisodeState
+from emote.utils.gamma_matrix import discount, make_gamma_matrix, split_rollouts
+from emote.utils.spaces import MDPSpace
+
+
+class GenericAgentProxy(AgentProxy):
```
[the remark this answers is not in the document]
jaxs-ribs: I fixed it :)

experiments/gym/train_dqn_cartpole.py
```diff
@@ -0,0 +1,304 @@
...
+    parser.add_argument("--use-wandb", action="store_true")
+    parser.add_argument(
+        "--wandb-run",
+        type=str,
+        default=None,
+        help="Short display name of run for the W&B UI. Randomly generated by default.",
+    )
+    arg = parser.parse_args()
+
+    env = DictGymWrapper(AsyncVectorEnv([_make_env() for _ in range(arg.num_envs)]))
```
singhblom, line 216: Could we break this out into a `main(args)`? I think that tends to look a bit cleaner.

emote/algorithms/dqn.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def begin_batch(self, next_observation, rewards, masks):
+        next_q_values = self.target_q_net(**next_observation)
+        max_next_q_values = next_q_values.max(1)[0].unsqueeze(1)
+        last_step_masks = split_rollouts(masks, self.rollout_len)[:, -1]
+        max_next_q_values = torch.multiply(max_next_q_values, last_step_masks)
+        scaled_reward = self.reward_scale * rewards
+        scaled_rewards = split_rollouts(scaled_reward, self.rollout_len).squeeze(2)
+        q_target = discount(scaled_rewards, max_next_q_values, self.gamma_matrix).detach()
+
```
singhblom, line 62: Maybe log a scalar or two?

singhblom (approved):
Looks good, only some minor comments. It would be nice to start thinking about documentation for this as well, this could definitely go into the "emote quick-start guide" that doesn't exist yet but probably should. With, like, a few graphs showing performance, a screenshot/gif of cartpole and the main part of the training loop shown.

emote/algorithms/generic_agent_proxy.py
```diff
@@ -0,0 +1,105 @@
...
+import torch
+
+from torch import nn
+
+from emote.proxies import AgentProxy
+from emote.typing import AgentId, DictObservation, DictResponse, EpisodeState
+from emote.utils.spaces import MDPSpace
+
+
+class GenericAgentProxy(AgentProxy):
```
AliGhadirzadeh, line 13: I do not think that GenericAgentProxy belongs to algorithms. I'd suggest to keep "emote/algorithms" for implementing methods such as SAC and DQN. And, putting generic codes into corresponding emote folders. As an example, we can have a python file for proxies and put this into the proxies folder.

AliGhadirzadeh (approved):
Nice work! I just left one comment regarding the structure of the code.

emote/algorithms/generic_agent_proxy.py
jaxs-ribs: Good point, changed that!

PR #176, opened 2023-10-10
Use !WARNING in markdown

singhblom:
I just learned about this new markdown element you can use on github and wanted to try it in the readme.


tgolsson:
Is this warning even true anymore though? We're using it in production... :P

singhblom:
> Is this warning even true anymore though? We're using it in production... :P

Haha. Noooo! The whole point of the PR is that I want to use the element. They also support IMPORTANT and NOTE. Can you think of anything that would warrant those?

PR #178, opened 2023-10-12
Implement Coverage Based Sampling Strategy

jaxs-ribs:
This PR adds a sampling strategy that prioritises samples that haven't been visited yet over samples that have. The chance of an episode being sampled is inversely proportional to the amount of times it has been sampled. This is additionally weighted by an `alpha` parameter, where a higher alpha means more weighting of less-sampled episodes. 

A test is also added to validate that recently added samples have a higher chance of being chosen.

### DIFF
--- a/emote/memory/coverage_based_strategy.py
+++ b/emote/memory/coverage_based_strategy.py
@@ -13,18 +13,18 @@ from .strategy import EjectionStrategy, SampleStrategy, Strategy
 
 
 class CoverageBasedStrategy(Strategy):
-    """A sampler intended to sample based on coverage of experiences, 
-    favoring less-visited states. This base class can be used for implementing 
+    """A sampler intended to sample based on coverage of experiences,
+    favoring less-visited states. This base class can be used for implementing
     various coverage-based sampling strategies."""
 
-    def __init__(self):
+    def __init__(self, alpha=0.5):
         super().__init__()
         self._identities = {}
         self._sample_count = {}
         self._ids = []
         self._prios = []
         self._dirty = False
-        self._alpha = 0.5
+        self._alpha = alpha
 
     def track(self, identity: int, sequence_length: int):
         self._dirty = True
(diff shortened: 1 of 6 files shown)


emote/memory/coverage_based_strategy.py
```diff
@@ -0,0 +1,110 @@
...
+        original_prios = np.array(tuple(self._identities.values())) / sum(self._identities.values())
+        self._ids = np.array(tuple(self._identities.keys()), dtype=np.int64)
+
+        sample_prios = np.array(
+            [1 / (self._sample_count[id] + 1) ** self._alpha for id in self._ids]
+        )
+        combined_prios = original_prios * sample_prios
+
+        sum_prios = sum(combined_prios)
+        self._prios = combined_prios / sum_prios
```
tgolsson, line 50: I *think* this needs to account for length of episodes somehow, something like `sample_count[i] / len[i]` as a weight? So we get uniformity over transitions, and not just episodes.

tgolsson (commented):
Nice!

I'm wondering if the sampler is too smart here, with a few years having passed since I wrote the original uniform sampling code. In particular, I'm wondering if this *is* modelling a sumtree. Or at least the uniform does. We just sample from it using another method. 

For example, in the uniform case we could just have a list of `[*[i]*len(i), *[i+1]*len(i+1), ...]` and pick random indexes, which would pretty much give you O(1) sampling without rebalancing. Needs to maintain and defrag a large buffer, but it seems cheaper than the rebalancing.

For this method it's a bit hairier, because your "length" of each slot is *inversely* proportional to the tracked metric. Wonder if that could be solved...

emote/memory/coverage_based_strategy.py
jaxs-ribs: Isn't that being done through `combined_prios`? The multiplication should take care of that 🤔

tgolsson: Right, yes.

tgolsson (approved):
Nice, LGTM!

tgolsson:
Maybe worth adding a changelog entry to this? We're awfully bad at it but it's quite useful :D

tgolsson (approved):
LGTM!

PR #180, opened 2023-10-23
Add Documentation for Cartpole Training

jaxs-ribs:
Basic documentation on how to run the cartpole test, what the arguments represent, and how to track metrics in tensorboard.


singhblom (approved):
Really nice! If you want to make it even better I think `train_dqn_cartpole.py` could benefit from many more comments so we gently walk the user through it, but that would just be icing. Good job.

jaxs-ribs:
> Really nice! If you want to make it even better I think `train_dqn_cartpole.py` could benefit from many more comments so we gently walk the user through it, but that would just be icing. Good job.

Will do that, good point!

PR #181, opened 2023-10-27
Add a joint memory loader capable of sampling data from multiple memory loaders

klashenriksson:
This PR adds a `JointMemoryLoader` object that can sample data from multiple different `MemoryLoader`s. This is needed for a lot of imitation related training.

### DIFF
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -464,15 +464,10 @@ class MemoryLoader:
 
 
 class JointMemoryLoader:
-    """A memory loader capable of loading data from multiple `MemoryLoader`s.
+    """A memory loader capable of loading data from multiple `MemoryLoader`s."""
 
-    If a datagroup is specified via the `data_group` param, it will place all loaded data in the specified datagroup,
-    otherwise it will directly return the data fetched from each individual loader.
-    """
-
-    def __init__(self, loaders: list[MemoryLoader], data_group: str | None = None):
+    def __init__(self, loaders: list[MemoryLoader]):
         self._loaders = loaders
-        self._data_group = data_group
 
     def is_ready(self):
         return all(loader.is_ready() for loader in self._loaders)
@@ -489,10 +484,18 @@ class JointMemoryLoader:
             for loader in self._loaders:
                 out.update(next(iter(loader)))
 
-            if self._data_group is not None:
-                out = {self._data_group: out}
-
             yield out
+
+
+class JointMemoryLoaderWithDataGroup(JointMemoryLoader):
+    """A JointMemoryLoader that places its data inside of a user-specified datagroup."""
+
+    def __init__(self, loaders: list[MemoryLoader], data_group: str):
+        super().__init__(loaders)
+        self._data_group = data_group
+
+    def __iter__(self):
+        yield {self._data_group: next(super().__iter__())}
 
 
 class MemoryWarmup(Callback):
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -475,7 +475,7 @@ class JointMemoryLoader:
         self._data_group = data_group
 
     def is_ready(self):
-        return all([loader.is_ready() for loader in self._loaders])
+        return all(loader.is_ready() for loader in self._loaders)
 
     def __iter__(self):
         if not self.is_ready():
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -471,11 +471,10 @@ class JointMemoryLoader:
         self._loaders = loaders
         self._size_key = size_key
 
-        datagroups = (loader.data_group for loader in loaders)
-        counts = collections.Counter(datagroups)
+        counts = collections.Counter((loader.data_group for loader in loaders))
         counts_over_1 = {k: count for k, count in counts.items() if count > 1}
         if len(counts_over_1) != 0:
-            raise Exception(
+            raise ValueError(
                 f"""JointMemoryLoader was provided MemoryLoaders that share the same datagroup. This will clobber the joint output data and is not allowed.
                 Here is a dict of each datagroup encountered more than once, and its occurance count: {counts_over_1}"""
             )
@@ -485,7 +484,7 @@ class JointMemoryLoader:
 
     def __iter__(self):
         if not self.is_ready():
-            raise Exception(
+            raise RuntimeError(
                 """memory loader(s) in JointMemoryLoader does not have enough data. Check `is_ready()`
                 before trying to iterate over data."""
             )
--- a/emote/memory/memory.py
+++ b/emote/memory/memory.py
@@ -467,8 +467,9 @@ class MemoryLoader:
 class JointMemoryLoader:
     """A memory loader capable of loading data from multiple `MemoryLoader`s."""
 
-    def __init__(self, loaders: list[MemoryLoader]):
+    def __init__(self, loaders: list[MemoryLoader], size_key: str = "batch_size"):
         self._loaders = loaders
+        self._size_key = size_key
 
         datagroups = (loader.data_group for loader in loaders)
         counts = collections.Counter(datagroups)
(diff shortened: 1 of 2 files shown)


emote/memory/memory.py
```diff
@@ -463,6 +463,38 @@ def __iter__(self):
...
+    If a datagroup is specified via the `data_group` param, it will place all loaded data in the specified datagroup,
+    otherwise it will directly return the data fetched from each individual loader.
+    """
+
+    def __init__(self, loaders: list[MemoryLoader], data_group: str | None = None):
+        self._loaders = loaders
+        self._data_group = data_group
+
+    def is_ready(self):
+        return all([loader.is_ready() for loader in self._loaders])
```
tgolsson, line 478: Nit: Should be fine to pass a generator instead of using a listcomp.

emote/memory/memory.py
```diff
@@ -463,6 +463,38 @@ def __iter__(self):
...
+                before trying to iterate over data."""
+            )
+
+        while True:
+            out = {}
+            for loader in self._loaders:
+                out.update(next(iter(loader)))
+
+            if self._data_group is not None:
+                out = {self._data_group: out}
```
tgolsson, line 493: Hmm, it feels like there's some risk of clobbering here... Maybe we can check some trivial errors at least? Ensure the inner data groups don't overlap? Otherwise ensure the keys don't overlap? Or is that something we want? Wondering how we can avoid future hidden errors here.

emote/memory/memory.py
```diff
@@ -463,6 +463,38 @@ def __iter__(self):
             yield {self.data_group: data, self.size_key: data[self.size_key]}
 
 
+class JointMemoryLoader:
+    """A memory loader capable of loading data from multiple `MemoryLoader`s.
+
+    If a datagroup is specified via the `data_group` param, it will place all loaded data in the specified datagroup,
+    otherwise it will directly return the data fetched from each individual loader.
```
tgolsson, line 470: Ok; hear me out. Subclasses. Default behavior is no data group, a subclass from this that just does the following:

```py
class JointMemoryLoaderWithDataGroup(JointMemoryLoader):
    def __init__(self, ...):
        super().__init__(self)
        self.data_group = ...

    def __iter__(self):
        return { self.data_group: next(super()) }
```
       
Each of these classes becomes more intentional!

tgolsson (changes requested):

emote/memory/memory.py
klashenriksson: Very nice, thanks. I did not think of it.

emote/memory/memory.py
klashenriksson: Yeah I think it makes sense to ensure the keys are unique.. Feels like something we only have to do once though, the keys should be static throughout training yes?

emote/memory/memory.py
klashenriksson: Do I smell inheritance?? ;) No but sure, I think I like that as well.

emote/memory/memory.py
singhblom: Ooof. I have no good insights except for saying that the size key has always felt out of place. I don't know what the right solution is and the context switch is too large for me to think hard about emote right now but if we could come up with a nicer structure here that would be really nice. If the interface for memory loaders doesn't work for union memory loaders it feels like we need to think about if we have the right abstraction.

tgolsson: I'm wondering whether the size should be part of the group instead of adjacent? I don't remember why we (I?) designed it to be adjacent. Maybe to keep the group pure with data from inside the memory...

tgolsson: That sounds fine; can you do a check for which one is used? I think with our decorators for automatic unpacking it'd work fine still, I'm curious if there's some callback that *doesn't* use datagroups that does read the batch size.

singhblom: Adjacent _and_ in sounds like the unclearest of all possible options.

klashenriksson: I will look into it. Some tests seem to rely on it but probably an easy fix.

klashenriksson: I think I found why it was placed outside in the first place. The `Trainer` tries to fill the callback state dict with `bp_samples` fetched from a user-defined `batch_size_key`, and it expects a callback to fill in that key:

```
def train(..):
    ...
    for bp_step, batch in zip(counter, self.dataloader):
        ....
        self.state["bp_samples"] += self.state[self._batch_size_key]
```

and that `bp_samples` entry is used for performance logging. If we want to keep this system I think we can solve the clobbering by only injecting a size key on the outermost dataloader. I have updated the joint data loader with this change, wdyt?

singhblom: To me there is something that feels weird about the way we've structured the data here. It feels like the batch size key is misplaced in some way. That said I think your updated code solves all our current problems neatly. Let's go with this and then think more about the problem next time we work on the memory.

tgolsson: Do we drive any schedules by bp_samples? If so, they're now much faster. :-)

singhblom: Wait? Why much faster? It seems to me like what Klas did should keep the speed the same?

singhblom (approved):

emote/memory/memory.py
tgolsson: Wouldn't it run by the sum of the batch sizes instead of whichever one was the last one before?

tgolsson: I'm not saying it's *wrong*, I was just pointing out that something depending on `bp_samples` progressing at a certain rate the rate would be much higher now.

emote/memory/memory.py
```diff
@@ -463,6 +464,58 @@ def __iter__(self):
...
+
+    def __init__(self, loaders: list[MemoryLoader], size_key: str = "batch_size"):
+        self._loaders = loaders
+        self._size_key = size_key
+
+        datagroups = (loader.data_group for loader in loaders)
+        counts = collections.Counter(datagroups)
+        counts_over_1 = {k: count for k, count in counts.items() if count > 1}
+        if len(counts_over_1) != 0:
+            raise Exception(
```
tgolsson, line 478: ValueError

emote/memory/memory.py
```diff
@@ -463,6 +464,58 @@ def __iter__(self):
...
+
+    def is_ready(self):
+        return all(loader.is_ready() for loader in self._loaders)
+
+    def __iter__(self):
+        if not self.is_ready():
+            raise Exception(
+                """memory loader(s) in JointMemoryLoader does not have enough data. Check `is_ready()`
+                before trying to iterate over data."""
+            )
```
tgolsson, line 491: RuntimeError

emote/memory/memory.py
```diff
@@ -463,6 +464,58 @@ def __iter__(self):
...
 
 
+class JointMemoryLoader:
+    """A memory loader capable of loading data from multiple `MemoryLoader`s."""
+
+    def __init__(self, loaders: list[MemoryLoader], size_key: str = "batch_size"):
+        self._loaders = loaders
+        self._size_key = size_key
+
+        datagroups = (loader.data_group for loader in loaders)
```
tgolsson, line 474: This seems quite ineffectual, can't you just put it in the Counter call?

tgolsson (changes requested):
Minor nitpicks about using a generic Exception and one pointless variable! Apart from that LGTM.

emote/memory/memory.py
klashenriksson: I did not think you were saying that! Yes we should probably address those before merging, I just mean that to me it makes sense to instead rely on bp_step or a particular loaders batch size.

emote/memory/memory.py
klashenriksson: Indeed!

AliGhadirzadeh:
@klashenriksson what is the status of this? Will it be possible to merge it soon?

klashenriksson:
> @klashenriksson what is the status of this? Will it be possible to merge it soon?

We just have to go through our internal stuff to make sure we dont have any schedules relying on `bp_samples`, and if so change so we rely on `bp_step`. I can have a look at this tomorrow. I took a quick look now and did not see anything.. But will have a more thorough look.

emote/memory/memory.py
klashenriksson: I've had a look and could not find anything relying on `bp_samples` so I think this is fine to merge.

tgolsson (approved):

PR #182, opened 2023-10-27
Add metrics logger for MemoryLoaders

klashenriksson:
This PR adds a simple callback that periodically logs metrics from `MemoryLoader`s. What do you think? Are you ok with the metric namings?


tgolsson:
Hmm, doesn't this overlap quite a bit with #179?

klashenriksson:
Aha! Missed that issue. Yeah then this pretty much is a full overlap. I guess we could use this for just the dataloader timings, but not sure how much new info that gives?

PR #184, opened 2023-11-06
Upgrade to Python 3.10

jaxs-ribs:
Upgrades emote from 3.9 to 3.10 the main changes are: 

- You can now combine with statements, so instead of 

```
with open('file1.txt') as file1:
    with open('file2.txt') as file2:
    	bla
```

You can use 

```
with (
    open('file1.txt') as file1,
    open('file2.txt') as file2
):
	bla
```

- Instead of using `Union`, you can now use `|`. So `name: Union[str, bytes]` to `name: str | bytes`. 
- `match` statements have been added, similar to rust, for datastructures

The first two points have been adjusted in this repo, grouping `with`s together where possible and using `|` instead of `Union`s. 

Pdm had to be version 2.5.6 to properly update the lock files, otherwise it wouldn't work.


singhblom (approved):
This looks good to me. Have you tried it with the nightlies etc?

tgolsson (approved):
LGTM!

PR #185, opened 2023-11-09
Add schedules

spetravic:
This PR adds a few schedules that can be used for adjusting parameters during training.

### DIFF
--- a/emote/extra/schedules.py
+++ b/emote/extra/schedules.py
@@ -1,3 +1,5 @@
+import math
+
 from dataclasses import dataclass
 
 from emote.utils.math import truncated_linear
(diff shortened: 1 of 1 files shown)


singhblom (approved):
Nice.

tgolsson:
Too late to the party; but "get_last_value" is a *terrible* name for the current value... ;-)

spetravic:
@tgolsson I wanted to keep it consistent with get_last_lr() used by Pytorch in their learning rate schedulers because according to them it returns the last computed value. Of course I can rename it :D

tgolsson:
My personal opinion is that we shouldn't copy other people's bad practice. :P I think just `value` or `current` would make a lot more sense! `last` has connotations of "the end", i.e, the value it'd be when the schedule ends. I'd be happy with a property too, as it makes it clear it's cheap but still immutable.
```
@property
def value(self):
    return self._current_val
```

PR #186, opened 2023-11-09
Update AlphaLoss to support entropy schedules

spetravic:
This PR adds support for entropy schedules in the AlphaLoss callback. Also I removed the multiplication term in the default target entropy value since it defaulted to a very small value (CC @AliGhadirzadeh) and now we have the option to set the target value directly instead of scaling the number of actions by changing `entropy_eps`.

### DIFF
--- a/emote/algorithms/sac.py
+++ b/emote/algorithms/sac.py
@@ -292,7 +292,7 @@ class AlphaLoss(LossCallback):
         with torch.no_grad():
             _, logp_pi = self.policy(**observation)
             entropy = -logp_pi
-            error = entropy - self.t_entropy.get_last_val()
+            error = entropy - self.t_entropy.value
         alpha_loss = torch.mean(self.ln_alpha * error.detach())
         assert alpha_loss.dim() == 0
         self.log_scalar("loss/alpha_loss", alpha_loss)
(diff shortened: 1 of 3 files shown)


AliGhadirzadeh (approved):
Thanks! It looks good to me.

emote/algorithms/sac.py
```diff
@@ -267,10 +267,10 @@ def __init__(
         lr_schedule: Optional[optim.lr_scheduler._LRScheduler] = None,
         n_actions: int,
         max_grad_norm: float = 10.0,
-        entropy_eps: float = 0.089,
         max_alpha: float = 0.2,
         name: str = "alpha",
         data_group: str = "default",
+        t_entropy: float | Schedule = None,
```
tgolsson, line 273: `float | Schedule | None = None` if it's optional. :P

tgolsson (changes requested):

emote/algorithms/sac.py
spetravic: Ahh types will be the death of me

singhblom (approved):

tgolsson:
Maybe a few tests for this? Not sure how much "collateral" is needed to make a test of this work... Would be nice to see what happens if you pass an invalid value f.ex.

tgolsson (approved):
Nice! 😍

PR #187, opened 2023-11-09
Rename get_last_val

spetravic:
Yet another reminder to self to wait for Tom's comments before merging :(

### DIFF
--- a/emote/extra/schedules.py
+++ b/emote/extra/schedules.py
@@ -25,10 +25,11 @@ class Schedule:
         self.steps = steps
 
         self._step_count = 0
-        self._last_val = initial
+        self._val = initial
 
-    def get_last_val(self) -> float:
-        return self._last_val
+    @property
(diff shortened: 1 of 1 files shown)


emote/extra/schedules.py
```diff
@@ -25,10 +25,11 @@ def __init__(self, initial: float, final: float, steps: int):
         self.steps = steps
 
         self._step_count = 0
-        self._last_val = initial
+        self._val = initial
```
singhblom, line 28: Hm. I don't actually like _val. Could we maybe go for something a bit more descriptive here? Since we call it `value_min` and `value_max` I think just `current_value` would be nicer.

singhblom (approved):
Small naming comment since we're discussing naming anyways, but I have no strong feelings here and am fine with this PR as is.

tgolsson (approved):

PR #188, opened 2023-11-09
Fix wrong `onnx.load_model` arguments

jaxs-ribs:
We called this with a second argument that got ignored in onnx 1.14.1, but with 1.15.0, this is not legal anymore. Removing the second argument should fix this.


tgolsson (approved):
Upstream issue filed here; https://github.com/onnx/onnx/issues/5746, fix LGTM ;-)

PR #189, opened 2023-11-10
fix bug in logging timers

tgolsson:
This PR:

* Adds an overrideable name to the memory TableMemoryProxy
* Exposes that name as a property
* Extends the proxy base to allowed reading properties from the inner classes
* Uses the above property for logging


spetravic (approved):
LGTM!

PR #191, opened 2023-11-20
Upgrade to PDM 2.10

tgolsson:
This upgrades to PDM 2.10 and `pdm-plugin-torch` to 23.4. As part of this I've removed both hooks - install now happens with `pdm install --plugins` and the lock will have to be explicit. This removes some unncessary magic and makes it more explicit. 

Will announce on Slack once merged.


.buildkite/validate-lockfile.sh
```diff
@@ -10,7 +10,7 @@ ${PDM_COMMAND:1:-1} torch lock --check || TORCH_EXIT_CODE=$?
 
 if [ $EXIT_CODE -ne 0 ] || [ $TORCH_EXIT_CODE -ne 0 ]; then
 	cat << EOF | buildkite-agent annotate --style "error" --context "lockfile"
-:lock: Lockfiles are outdated. Please run \`pdm lock\` and commit the result.
+:lock: Lockfiles are outdated. Please run \`pdm lock && pdm torch lock\` and commit the result.
```
jaxs-ribs, line 13: Is this change related to changing the version?

jaxs-ribs (approved):
LGTM!

.buildkite/validate-lockfile.sh
tgolsson: Related to this: > the lock will have to be explicit

We can reinstate this once https://github.com/pdm-project/pdm/pull/2408 is released.

tgolsson: The reason is that we can't chain them here properly since our CI images have lots of PDMs but none actually called `pdm`. So then we have to disable hooks in CI which is a pain.

PR #192, opened 2023-11-20
Update Github Workflows for Python3.10

jaxs-ribs:
Since updating to python3.10, github actions failed as specifiying the python-version in the workflow files with only 3.10 got only parsed as 3.1. Setting this to 3.10.4 makes it find a proper build. 

This PR also separates the build-docs and deploy jobs, and runs the former whenever a PR is opened.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,9 +16,6 @@ concurrency:
 
 jobs:
   build-docs:
-    environment:
-        name: github-pages
-        url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


tgolsson (approved):
Nice, let's see if it publishes correctly on main as well! Always a pain to discover the conditions are wrong somewhere.

.github/workflows/docs.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Setup PDM
         uses: pdm-project/setup-pdm@v3
         with:
-          python-version: 3.10.4
+          python-version: '3.10'
```
singhblom, line 33: lol yaml.

singhblom (approved):

PR #195, opened 2024-01-31
Ali/amp

AliGhadirzadeh:
AMP implementation

### DIFF
--- a/emote/algorithms/amp.py
+++ b/emote/algorithms/amp.py
@@ -39,7 +39,8 @@ class DiscriminatorLoss(LossCallback):
     def __init__(
         self,
         discriminator: nn.Module,
-        state_map_fn: Callable[[Tensor], Tensor],
+        imitation_state_map_fn: Callable[[Tensor], Tensor],
+        policy_state_map_fn: Callable[[Tensor], Tensor],
         grad_loss_weight: float,
         optimizer: torch.optim.Optimizer,
         lr_schedule: torch.optim.lr_scheduler._LRScheduler,
@@ -56,7 +57,8 @@ class DiscriminatorLoss(LossCallback):
             data_group=None,
         )
         self._discriminator = discriminator
-        self._state_map_function = state_map_fn
+        self._imitation_state_map_function = imitation_state_map_fn
+        self._policy_state_map_function = policy_state_map_fn
         self._grad_loss_weight = grad_loss_weight
         self._obs_key = input_key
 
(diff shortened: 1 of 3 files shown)


emote/algorithms/amp.py
```diff
@@ -0,0 +1,182 @@
...
+class DiscriminatorLoss(LossCallback):
+    """
+    This loss is used to train a discriminator for
+    adversarial training.
+    """
+
+    def __init__(
+        self,
+        discriminator: nn.Module,
+        state_map_fn: Callable[[Tensor], Tensor],
```
klashenriksson, line 42: I think we should split this up into imitation_state_map_fn and policy_state_map_fn, as they might require different mappings (this is the case for my work for instance)

emote/algorithms/amp.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        imitation_batch_size = imitation_batch["batch_size"]
+        policy_data_batch_size = policy_batch["batch_size"]
+
+        pos_obs: Tensor = imitation_batch["observation"][self._obs_key]
+        pos_next_obs: Tensor = imitation_batch["next_observation"][self._obs_key]
+        neg_obs: Tensor = policy_batch["observation"][self._obs_key]
+        neg_next_obs: Tensor = policy_batch["next_observation"][self._obs_key]
+
+        (pos_obs,) = (self._state_map_function(pos_obs),)
```
klashenriksson, line 83: why the tuple wrap?

emote/memory/memory.py
```diff
@@ -486,6 +486,30 @@ def __iter__(self):
             yield {self.data_group: data, self.size_key: data[self.size_key]}
 
 
+class MemoryLoaderList:
```
klashenriksson, line 489: Please see my pull request here: https://github.com/EmbarkStudios/emote/pull/181 . The current implementation risks clobbering data if multiple loaders have the same datagroup, and there is also some troubles with the size_key that each loader adds.

klashenriksson (changes requested):

emote/memory/memory.py
AliGhadirzadeh: Nice! I removed it from my PR then :)

emote/algorithms/amp.py
AliGhadirzadeh: Have no idea how this happened! It should be some automatic code formatting. Thanks for the catch :)

emote/algorithms/amp.py
AliGhadirzadeh: Fixed :)

emote/algorithms/amp.py
```diff
@@ -0,0 +1,184 @@
...
+        discriminator: nn.Module,
+        state_map_fn: Callable[[Tensor], Tensor],
+        style_reward_weight: float,
+        rollout_length: int,
+        observation_key: str,
+        data_group: str,
+    ):
+        super().__init__()
+        self._discriminator = discriminator
+        self._order = 0
```
klashenriksson, line 134: Can this become a problem when paired with SAC? If QTarget runs before this callback it will miss the amp reward. Perhaps order just be an input parameter?

klashenriksson: What happens if QTarget also has order=0?

klashenriksson: Alright, I think this is fine then for this PR, users will have to assert that it runs before QTarget (or any other relevant callback).

I think I prefer to have the constructors take in an `order` parameter rather than having a `set_order` method. The callbacks are sorted by their order only once when given to the `Trainer`, so if you would call `set_order` at a later time, it would not have any effect on the actual training. By only exposing it during construction makes it clear that this is a "set once, never change" thing. A `get_order` method sounds nice though, we could make it a `@property`.

Anyway, that sounds like a different PR, approving this, nice stuff!

klashenriksson (approved):

singhblom (approved):

PR #197, opened 2024-03-26
Collect cum logs under their own heading

singhblom:
We rarely need the cumulative logs, but when they are needed it's really good to have. I have therefore hidden them all under a common `cumulative/` heading in tb.

### DIFF
--- a/emote/callbacks/logging.py
+++ b/emote/callbacks/logging.py
@@ -53,7 +53,7 @@ class TensorboardLogger(Callback):
 
             for k, v in cb.windowed_scalar_cumulative.items():
                 k_split = k.split("/")
-                k_split[0] = k_split[0] + "_" + suffix
+                k_split[0] = f"cumulative/{k_split[0]}_{suffix}"
                 k = "/".join(k_split)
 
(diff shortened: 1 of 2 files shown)


tgolsson:
Won't this be `cumulative/abc/cumulative`? Seems redundant.

emote/callbacks/logging.py
```diff
@@ -53,7 +53,7 @@ def end_cycle(self, bp_step, bp_samples):
 
             for k, v in cb.windowed_scalar_cumulative.items():
                 k_split = k.split("/")
-                k_split[0] = k_split[0] + "_" + suffix
+                k_split[0] = f"cumulative/{k_split[0]}_{suffix}"
                 k = "/".join(k_split)
 
                 self._writer.add_scalar(f"{k}/cumulative", v, bp_step)
```
spetravic, line 59: We could do "cumulative/{k}" when adding the scalar instead to avoid the duplication

singhblom: I made it redundant so it would be easy to exclude when you search for it. But I am happy to skip it.

PR #198, opened 2024-04-12
Add NextNElementWrapper virtual storage

klashenriksson:
This PR adds a small virtual storage wrapper that samples the next `N` elements ahead. This is needed for AMP training where the policy does not execute actions at the same frequency as the imitation reference.

### DIFF
--- a/emote/memory/storage.py
+++ b/emote/memory/storage.py
@@ -202,6 +202,48 @@ class NextElementMapper(VirtualStorage):
         return NextElementMapper(storage, shape, dtype, only_last=True)
 
 
+class NextNElementWrapper(VirtualStorage):
+    """Simple mapper that can be used to sample a specified N steps over"""
+
+    class Wrapper:
+        def __init__(self, item, n: int):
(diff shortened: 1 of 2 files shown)


emote/memory/storage.py
```diff
@@ -202,6 +202,48 @@ def with_only_last(storage, shape, dtype):
...
+            self._item = item
+            self._n = n
+
+        def __getitem__(self, key):
+            if isinstance(key, int):
+                key += self._n
+            elif isinstance(key, tuple):
+                key = tuple(k + self._n for k in key)
+            elif isinstance(key, slice):
+                key = slice(key.start + self._n, key.stop + self._n, key.step)
```
singhblom, line 219: Maybe raise here if it's not any of these?

singhblom (approved):
Nice and clean. LGTM.

emote/memory/storage.py
klashenriksson: Sure!

